PR #1, opened 2016-02-25
More problems

vepiteski:
A first batch of new problems... I consider regrouping problems in thematic subfolders.


sarvjeetsinghghotra:
@dpo Can you please give me some small task to start contributing to this package.

dpo:
@sarvjeetsinghghotra Thanks for asking! I would say that at the moment, what we need is more problems and a way to classify/search/select problems with certain features. An initial small task would be to add a few unconstrained problems from the literature. For example, we've been adding problems from [this paper](http://www.cs.cas.cz/matonoha/download/V1081.pdf) little by little.

dpo:
I merged almost all of this. I didn't merge:

* the dummy arguments given to `rosenbrock()` and `woods()`: I'd like to find a longer-term solution such as problem classification.
* the tests that involve `NLPModels`: they currently fail because `MathProgBase` has changed.

Thank you!

vepiteski:
Le 2016-03-05 00:48, Dominique a écrit :
>
> I merged almost all of this. I didn't merge:
>
>   * the dummy arguments given to |rosenbrock()| and |woods()|: I'd
>     like to find a longer-term solution such as problem classification.
>   * the tests that involve |NLPModels|: they currently fail because
>     |MathProgBase| has changed.
>

Salut Dominique,

La branche que tu as créée, new_mpb  doit bien marcher, non? J'ai 
dupliqué les changements et ça marche sur mon setup.

Bon, j'ai changé la branche, puis la nouvelle version de MPB et de JuMP 
s'est installée, et mon code déconne, des NaN apparaissent ... Il semble 
que certains problèmes sont épargnés, mais la version actuelle semble 
super instable...

Je suis bloqué donc pour l'instant. Saurais-tu comment revenir à la 
précédente version en attendant que la nouvelle soit plus stable?

JPD

PR #8, opened 2016-10-03
Implementation of the models 111 and 112 from HS

nicoleantunes:


src/hs111.jl
```diff
@@ -0,0 +1,54 @@
...
+
+  nlp = Model()
+
+  for j=1:10
+    x0[j] = -2.3
+  end
+
+  c = [-6.089, -17.164, -34.054, -5.914, -24.721, -14.986, -24.100, -10.708, -26.662, -22.179]
+
+  @variable(nlp, x[i=1:10], start=x0[i])
```
dpo: Wouldn't it be simpler to write
```julia
@variable(nlp, x[i=1:10], start=-2.3)
```

dpo:
Thank you!

src/hs111.jl
nicoleantunes: Sure! I didn't see that... Already made a pull request with the changes.

dpo:
Last detail: the two files should be included from `src/OptimizationProblems.jl`. If you look at [the test results](https://travis-ci.org/JuliaSmoothOptimizers/OptimizationProblems.jl/jobs/164814627#L198), they're not currently being read.

nicoleantunes:
Sorry, I didn't know that. I have just updated [src/OptimizationProblems.jl ](https://github.com/nicoleantunes/OptimizationProblems.jl)with the includes of hs111 and hs112.

dpo:
Hm. It seems my previous comment has disappeared or never appeared here. When checking the HS book, it seems to me that both problems impose bounds on the variables.

nicoleantunes:
Bounds included.

src/hs111.jl
```diff
@@ -20,7 +20,7 @@ function hs111(args...)
 
   c = [-6.089, -17.164, -34.054, -5.914, -24.721, -14.986, -24.100, -10.708, -26.662, -22.179]
 
-  @variable(nlp, x[i=1:10], start=-2.3)
+  @variable(nlp, x[i=1:10], lowerbound = -100, start = -2.3, upperbound = 100)
```
dpo: Sorry to be difficult, but for such simple bounds it would be more readable to write
```julia
@variable(nlp, -100 ≤ x[i=1:10] ≤ 100, start = -2.3)
```
See http://www.juliaopt.org/JuMP.jl/0.14/refvariable.html#variables. Does that work?

src/hs111.jl
```diff
@@ -0,0 +1,50 @@
...
+export hs111
+
+"HS111 model"
+function hs111(args...)
+
+  nlp = Model()
+
+  c = [-6.089, -17.164, -34.054, -5.914, -24.721, -14.986, -24.100, -10.708, -26.662, -22.179]
+
+  @variable(nlp, x[i=1:10], lowerbound = -100, start = -2.3, upperbound = 100)
```
dpo: In this case, woulnd't it be more readable to write `@variable(nlp, -100 ≤ x[i=1:10] ≤ 100, start=-2.3)`? I think JuMP allows such compact syntax (or something close to it).

src/hs111.jl
```diff
@@ -20,7 +20,7 @@ function hs111(args...)
 
   c = [-6.089, -17.164, -34.054, -5.914, -24.721, -14.986, -24.100, -10.708, -26.662, -22.179]
 
-  @variable(nlp, -100 < x[i=1:10] < 100, start = -2.3)
+  @variable(nlp, -100 =< x[i=1:10] =< 100, start = -2.3)
```
dpo: Does that work? Normally we write `<=` or `≤`.

src/hs1.jl
```diff
@@ -0,0 +1,37 @@
+# Hock and Schittkowski problem number 1.
+#
+#   Source:
+#   Problem 1 in
+#   W. Hock and K. Schittkowski,
+#   Test examples for nonlinear programming codes,
+#   Lectures Notes in Economics and Mathematical Systems 187,
+#   Springer Verlag, Heidelberg, 1981.
+#
+#   classification QQR2-AN-2-1
```
dpo: The classification strings are not the same as those in CUTEst. You can find the right ones in the corresponding SIF file. Sorry I didn't check that earlier.

dpo:
@nicoleantunes Many thanks for adding all those problems! Sorry to keep bothering you with small details.

nicoleantunes:
I guess they're all set now, but if still need changes, please let me now.

dpo:
I rearranged your commits to keep a clean history. In the future, please only group related changes under a single commit. `git rebase` is also a useful command to learn. Many thanks for your contribution!

PR #11, opened 2016-11-04
Changing classification of hs26.jl

Cebola94:
Dominic, sorry about that. Now it's ok.
Thank you.


dpo:
I don't understand... you changed 19 files to fix a classification string ;-).

dpo:
I think you forgot to pull the latest master. Then you can create a new branch off of it.

src/beale.jl
```diff
@@ -22,7 +22,7 @@ function beale(args...)
     @NLobjective(
       nlp,
       Min,
-      (-1.5 + x[1] * (1.0 - x[2]))^2 + (-2.25 + x[1] * (1.0 - x[2]^2))^2 + (-2.625 + x[1] * (1.0 - x[2]^3))^2
+      (1.5 + x[1] * (1.0 - x[2]))^2 + (2.25 + x[1] * (1.0 - x[2]^2))^2 + (2.625 + x[1] * (1.0 - x[2]^3))^2
```
dpo: It was right before, wasn't it? It would be easier if you wrote it exactly as in the paper.

PR #16, opened 2017-05-19
More convex models

dpo:


src/palmer1d.jl
```diff
@@ -0,0 +1,42 @@
...
+  Y = [78.596218, 65.77963,  43.96947,  27.038816, 14.6126,     6.2614,    1.538330,
+        0.000000,  1.188045,  4.6841,   16.9321,   33.6988,    52.3664,   70.1630,
+       83.4221,   88.3995,   78.596218, 65.77963,  43.96947,   27.038816, 14.6126,
+        6.2614,    1.538330,  0.000000,  1.188045,  4.6841,    16.9321,   33.6988,
+       52.3664,   70.1630,   83.4221,  108.18086,  92.733676, 108.18086,  92.733676]
+
+  @NLobjective(
+    nlp,
+    Min,
+    sum((Y[i] - (x[1] + x[2] * (X[i]^2) + x[3] * (X[i]^4) + x[4] * (X[i]^6) + x[5] * (X[i]^8) + x[6] * (X[i]^10) + x[7] * (X[i]^12)))^2 for i=1:35)
```
abelsiqueira: This can be changed to
```
sum((Y[i] - sum(x[j] * X[j]^(2j-2) for j = 1:7))^2 for i = 1:35)
```
I missed it in the last palmer problem, and it can also be changed in the others.
Note: `0^0 = 1` in Julia.

src/palmer1d.jl
```diff
@@ -0,0 +1,42 @@
...
+       -1.218612, -1.134464, -1.047198, -0.872665,  -0.698132, -0.523599, -0.349066,
+       -0.174533,  0.0000000, 1.788963,  1.745329,   1.658063,  1.570796,  1.483530,
+        1.396263,  1.308997,  1.218612,  1.134464,   1.047198,  0.872665,  0.698132,
+        0.523599,  0.349066,  0.174533, -1.8762289, -1.8325957, 1.8762289, 1.8325957]
+
+  Y = [78.596218, 65.77963,  43.96947,  27.038816, 14.6126,     6.2614,    1.538330,
+        0.000000,  1.188045,  4.6841,   16.9321,   33.6988,    52.3664,   70.1630,
+       83.4221,   88.3995,   78.596218, 65.77963,  43.96947,   27.038816, 14.6126,
+        6.2614,    1.538330,  0.000000,  1.188045,  4.6841,    16.9321,   33.6988,
+       52.3664,   70.1630,   83.4221,  108.18086,  92.733676, 108.18086,  92.733676]
```
abelsiqueira: Disclaimer: did not check all these values. Hoping that a copy-paste was done from a correct source.

src/palmer5c.jl
```diff
@@ -0,0 +1,50 @@
...
+    t[k, 2] = (2 * X[k] - a - b) / d
+    for l = 3:15
+      t[k, l] = 2 * t[k, l-1] * (2 * X[k] - a - b) / d - t[k, l-2]
+    end
+  end
+
+  @NLobjective(
+    nlp,
+    Min,
+    sum((Y[i] - (x[1] * t[i,1] + x[2] * t[i,3] + x[3] * t[i,5] + x[4] * t[i,7] + x[5] * t[i,9] + x[6] * t[i,11]))^2 for i=1:12)
```
abelsiqueira: Similar `sum` trick should work.

src/palmer5c.jl
```diff
@@ -0,0 +1,50 @@
...
+#
+#   model: H-N=C=Se TZVP+MP2
+#   fitting Y to A0 T_0 + A2 T_2 + A4 T_4 + A6 T_6 + A8 T_8 +
+#                A10 T_10 + A12 T_12 + A14 T_14
+#   where T_i is the i-th (shifted) Chebyshev polynomial
+#
+#   Source:
+#   M. Palmer, Edinburgh, private communication.
+#
+#   classification SUR2-RN-6-0
```
abelsiqueira: This should be `QUR2-RN-6-0`, but it is wrong on `MASTSIF` also.

src/palmer5d.jl
```diff
@@ -0,0 +1,35 @@
+#   A linear least squares problem
+#   arising from chemical kinetics.
+#
+#   model: H-N=C=Se TZVP+MP2
+#   fitting Y to to A0 + A2 X**2 + A4 X**4 + A6 X**6
+#
+#   Source:
+#   M. Palmer, Edinburgh, private communication.
+#
+#   classification SBR2-RN-4-0
```
abelsiqueira: This should be QU.

src/palmer6c.jl
```diff
@@ -0,0 +1,36 @@
...
+#   arising from chemical kinetics.
+#
+#   model: H-N=C=Se TZVP+MP2
+#   fitting Y to A0 + A2 X**2 + A4 X**4 + A6 X**6 + A8 X**8 +
+#                A10 X**10 + A12 X**12 + A14 X**14
+#
+#   Source:
+#   M. Palmer, Edinburgh, private communication.
+#
+#   classification SUR2-RN-8-0
```
abelsiqueira: This should be Q

src/palmer7c.jl
```diff
@@ -0,0 +1,36 @@
...
+#   arising from chemical kinetics.
+#
+#   model: H-N=C=Se TZVP+MP2
+#   fitting Y to A0 + A2 X**2 + A4 X**4 + A6 X**6 + A8 X**8 +
+#                A10 X**10 + A12 X**12 + A14 X**14
+#
+#   Source:
+#   M. Palmer, Edinburgh, private communication.
+#
+#   classification SUR2-RN-8-0
```
abelsiqueira: Q

src/palmer8c.jl
```diff
@@ -0,0 +1,36 @@
...
+#   arising from chemical kinetics.
+#
+#   model: H-N=C=Se TZVP+MP2
+#   fitting Y to A0 + A2 X**2 + A4 X**4 + A6 X**6 + A8 X**8 +
+#                A10 X**10 + A12 X**12 + A14 X**14
+#
+#   Source:
+#   M. Palmer, Edinburgh, private communication.
+#
+#   classification SUR2-RN-8-0
```
abelsiqueira: Q

src/power.jl
```diff
@@ -0,0 +1,28 @@
...
+function power(n::Int=1000)
+
+  nlp = Model()
+
+  @variable(nlp, x[i=1:n], start=1.0)
+
+  @NLobjective(
+    nlp,
+    Min,
+    sum((i * x[i]^2) for i=1:n)
```
abelsiqueira: Need a power 2 here: `sum(i * x[i]^2 for i = 1:n)^2` (assuming it's the quartic function from Oren). The AMPL model I'm following is also wrong here, but MASTSIF appears to be correct.

src/vardim.jl
```diff
@@ -0,0 +1,31 @@
...
+#   It Hessian matrix is dense.
+#
+#   Source:  problem 25 in
+#   J.J. More', B.S. Garbow and K.E. Hillstrom,
+#   "Testing Unconstrained Optimization Software",
+#   ACM Transactions on Mathematical Software, vol. 7(1), pp. 17-41, 1981.
+#
+#   See also Buckley#72 (p.98).
+#
+#   classification  OUR2-AN-V-0
```
abelsiqueira: This should be S.

src/cliff.jl
```diff
@@ -0,0 +1,30 @@
+#   The "cliff problem" in 2 variables
+#
+#   Source:  problem 206 (p. 46) in
+#   A.R. Buckley,
+#   "Test functions for unconstrained minimization",
+#   TR 1989CS-3, Mathematics, statistics and computing centre,
+#   Dalhousie University, Halifax (CDN), 1989.
```
dpo: We haven't found it either but it's referenced everywhere. Instead I found https://doi.org/10.1145/138351.138378, which might be the published version (not sure).

src/palmer1d.jl
dpo: Me too. I think @DahitoMA copy-pasted from the AMPL models.

dpo:
Many thanks for the thorough checks, @abelsiqueira !

dpo:
Here are updates. Several `palmer` problems share the same objective. Should we find a way to not repeat it and reuse it to minimize bugs?

abelsiqueira:
I think creating `palmerxx.data` files and a function
```
function palmer_ls(data :: String)
  if isfile("palmer$data.data")
```
and then specialized functions
```
palmer6c() = palmer_ls("6c")
```
could work.

abelsiqueira:
By the way, I didn't merge in case you want to make the reusing modification first. But it might be better to merge before and make another PR with the changes. Let me know.

dpo:
Thank @abelsiqueira. I agree.

PR #18, opened 2017-06-19
Convex problems

DahitoMA:
add more convex problems


dpo:
Thank you!

The comments in the `clplate` problems should indicate that you have eliminated fixed variables. Also, the `clplate` problems should be included in `src/OptimizationProblems.jl`.

PR #20, opened 2017-07-25
One dimensional problems

Goysa2:
Here are 28 one-dimensional optimization problems.


src/AMPGO02.jl
```diff
@@ -0,0 +1,23 @@
...
+# Problem 2 in
+# http://infinity77.net/global_optimization/test_functions_1d.html#d-test-functions
+# Andrea Gavana
+#
+# S. Goyette, Sherbrooke 2016/2017
+
+export AMPGO02
+
+function AMPGO02(n::Int64=1)
+   nlp = Model()
```
dpo: `n` should not be a parameter that the user can change if the problem size is fixed.

src/AMPGO03.jl
```diff
@@ -0,0 +1,23 @@
...
+
+function AMPGO03(n::Int64=1)
+   nlp = Model()
+
+   @variable(nlp, x, start=-10.0)
+
+   @NLobjective(
+    nlp,
+    Min,
+    -(1*sin(2*x+1)+2*sin(3*x+2)+3*sin(4*x+3)+4*sin(5*x+4)+5*sin(6*x+5)+6*sin(7*x+6))
```
dpo: Please use spaces between arithmetic operators for readability (à la Python). Please also be consistent across models (some use spaces and some don't). Thank you.

src/AMPGO07.jl
```diff
@@ -0,0 +1,26 @@
...
+
+   @NLobjective(
+    nlp,
+    Min,
+    #ifelse(x<=0.0,Inf,x)
+    ifelse(x<=0.0,Inf,sin(x)+sin((10/3)*x)+log(abs(x))-0.84*x+3)
+   )
+
+
+
```
dpo: Spurious blank lines.

src/cliff.jl
```diff
@@ -11,7 +11,7 @@
 export cliff
 
 "The 'cliff problem' in 2 variables"
-function cliff()
+function cliff(n::Int64=2)
```
dpo: `n` is fixed, so we don't want the user to specify it.

Goysa2:
I've corrected the syntax !

src/AMPGO02.jl
```diff
@@ -0,0 +1,23 @@
...
+
+function AMPGO02()
+   nlp = Model()
+
+   @variable(nlp, x, start=2.7)
+
+   @NLobjective(
+    nlp,
+    Min,
+    sin(x) + sin((10.0/3.0) * x)
```
dpo: Still a few spaces missing here and there.

dpo:
Thank you for all this! All other problems have a docstring. What could we use here?

Goysa2:
I've re-checked for missing spaces!
For the docstrings I've indicated that the problems are univariate and either unimodal and multimodal as well as the name of the problem

src/AMPGO02.jl
dpo: here

src/AMPGO04.jl
```diff
@@ -0,0 +1,24 @@
...
+"Univariate multimodal minimization problem AMPGO04"
+function AMPGO04()
+   nlp = Model()
+
+   @variable(nlp, x, start=1.9)
+
+   @NLobjective(
+    nlp,
+    Min,
+    -(16 * x^2 -24 * x + 5) * exp(-x)
```
dpo: here

src/AMPGO07.jl
```diff
@@ -0,0 +1,24 @@
...
+"Univariate multimodal minimization problem AMPGO07"
+function AMPGO07()
+   nlp = Model()
+
+   @variable(nlp, x, start=2.7)
+
+   @NLobjective(
+    nlp,
+    Min,
+    ifelse(x <= 0.0, Inf, sin(x) + sin((10/3) * x) + log(abs(x)) - 0.84 * x + 3)
```
dpo: here, and perhaps at other places...

Goysa2:
I've separated the fractions (10.0/3.0 is now 10.0 / 3.0). And I've broken lines that are too long.

dpo:
Thank you!

PR #22, opened 2017-07-31
Uniformize size for batch tests

vepiteski:
Added (args...) to fixed size problems. Changed the spec so that the n passed as parameter is indeed the size of the problem for clplate*.jl

### DIFF
--- a/src/clplatea.jl
+++ b/src/clplatea.jl
@@ -25,9 +25,14 @@
 export clplatea
 
 "The clamped plate problem (Strang, Nocedal, Dax)."
-function clplatea(p::Int=71, wght::Float64=-0.1)
+function clplatea(n::Int=5041, wght::Float64=-0.1)
 
-  nlp = Model()
+    p = floor(Int, sqrt(n))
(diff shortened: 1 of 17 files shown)


src/clplatea.jl
```diff
@@ -25,9 +25,14 @@
 export clplatea
 
 "The clamped plate problem (Strang, Nocedal, Dax)."
-function clplatea(p::Int=71, wght::Float64=-0.1)
+function clplatea(n::Int=5041, wght::Float64=-0.1)
 
-  nlp = Model()
+    p = floor(Int, sqrt(n))
+    p*p != n && warn("clplatea: number of variables adjusted from $n down to $(p*p) be square")
```
dpo: Let's remove those "be square". I hadn't noticed them before and I can't tell what they mean.

vepiteski: Oups, "in order for n to be a square integer". Ok, may be removed!

dpo: Could you please make those changes?

PR #26, opened 2018-05-01
New nonlinear least-squares problem NZF1

Xx-fractale-xX:


src/NZF1.jl
```diff
@@ -0,0 +1,29 @@
+# A. Montoison, Montreal, 04/2018.
```
dpo: Would you mind adding "Philippe Toint (private communication)" to indicate the provenance?

src/NZF1.jl
```diff
@@ -0,0 +1,29 @@
...
+
+export NZF1
+
+function NZF1(n :: Int=26)
+  
+  mod(n,13) != 0 && error(": number of variables must divisible by 13")
```
dpo: Instead of throwing an error, we prefer a warning, and we adjust `n` to the nearest multiple of 13.

abelsiqueira:
Being in JuMP, it's in the right place, but you could make an `ADNLPModel` version on NLSProblems also.

src/NZF1.jl
```diff
@@ -0,0 +1,31 @@
+# "Philippe Toint (private communication)"
+
+# A. Montoison, Montreal, 04/2018.
+
+export NZF1
+
+function NZF1(n :: Int=13)
+  
+  mod(n,13) != 0 && warn(": number of variables must divisible by 13")
```
dpo: Right, but in this case we also change `n` to the nearest multiple of 13.

src/NZF1.jl
dpo: We also need the classification string. See http://www.cuter.rl.ac.uk/Problems/classification.shtml. It would be something like SUR2-AN-V-0.

Xx-fractale-xX: Thanks for the link. I also found SUR2-AN-V-0 for the classification. I added it with the last commit.

dpo:
Thank you!

PR #27, opened 2018-05-01
Problems from the Schittkowski collection

Xx-fractale-xX:


src/CQRT11.jl
```diff
@@ -0,0 +1,28 @@
+#   Source: http://www.ai7.uni-bayreuth.de/test_problem_coll.pdf
+#   Problem 8 (p.31) in
+#   K. Schittkowski,
+#   "Test Examples for Nonlinear Programming Codes",
+#   Department of Computer Science, University of Bayreuth,
+#   D - 95440 Bayreuth, 2009.
+#
+# A. Montoison and L. Vincent, Montreal, 01/2018.
+
+export CQRT11
```
dpo: This is not the problem name. This is the classification string (which should appear in the comments, as it is stated in the PDF). The problem name is HS8, but it is already present in this collection: https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/master/src/hs8.jl

Xx-fractale-xX:
Yes, it's done.

Xx-fractale-xX:
I compared all models with an AMPL version and verify also myself when it was necessary. For models between 85 and 119 the verification was more difficult with an AMPL model, even if I verify many times, a mistake should still be present...

I deleted ```hs42.jl```, ```hs67.jl```, ```hs68.jl```, ```hs69.jl```, ```hs85.jl```, ```hs87.jl```, ```hs88.jl```, ```hs89.jl```, ```hs90.jl```, ```hs91.jl```, ```hs92.jl``` and ```hs99.jl``` because they used user-defined functions and automatic differentiation have some problems with them. 

I also updated ```OptimizationProblems.jl``` file to take into account new problems without the need to add their names manually in it.

dpo:
Thank you!!!

PR #41, opened 2019-10-27
add_some_cute_problems_10-2019

RenanOD:
Include problems (zangwil2, zecevic2, zecevic3, zecevic4, zy2), all from the CUTEr/st set.

The objective values agree with the references by solving them using JuMP with Ipopt.Solver.

I couldn't run compare_cutest.jl though due to a version requirement conflict between JuMP and NLPModelsJuMP.


RenanOD:
Because I couldn't find the original papers I used http://www.cuter.rl.ac.uk/Problems/mastsif.shtml as reference.

dpo:
@RenanOD I shared a Dropbox with you containing, among others, Buckley's report. You'll find the analytical form of the problems, the solution, etc.

src/zangwil2.jl
```diff
@@ -0,0 +1,31 @@
...
+    nlp = Model()
+
+    x0 = [3.0, 8.0]
+
+    @variable(nlp, x[i=1:2], start=x0[i])
+
+    @NLobjective(
+      nlp,
+      Min,
+      (-56*x[1] - 256*x[2] + 991 + 16*x[1]^2 + 16*x[2]^2 - 8*x[1]*x[2]) / 15
```
dpo: Maybe write this objective the way Buckley did?

dpo:
You can find AMPL models here: https://github.com/mpf/Optimization-Test-Problems/tree/master/cute. It's easier to check visually that the expressions agree, and you can use AmplNLReader to compare the models (there's a free version of AMPL for problems with less than 300 variables and 300 constraints).

src/zecevic2.jl
```diff
@@ -0,0 +1,54 @@
...
+#   Advanced Business School, Belgrade,
+#   (whatever is left of) Yugoslavia.
+#
+#   SIF input: Nick Gould, April 1993.
+#
+#   classification QLR2-AN-2-2
+#
+# R. O. Domingues, Paraná 10/2019.
+
+export zecevic2
```
dpo: Add blank line here

src/zecevic2.jl
```diff
@@ -0,0 +1,54 @@
...
+    nlp = Model()
+
+    x0 = [0.1, -0.1]
+
+    @variable(nlp, x[i=1:2], start=x0[i])
+
+    @NLobjective(
+      nlp,
+      Min,
+      2*x[2]^2 - 2*x[1] - 3*x[2]
```
dpo: Space around * (here and elsewhere)

src/zecevic3.jl
```diff
@@ -0,0 +1,54 @@
...
+#   Advanced Business School, Belgrade,
+#   (whatever is left of) Yugoslavia.
+#
+#   SIF input: Nick Gould, April 1993.
+#
+#   classification QQR2-AN-2-2
+#
+# R. O. Domingues, Paraná 10/2019.
+
+export zecevic3
```
dpo: Blank line

RenanOD:
Made the corrections, just didn't test with AmplNLReader due to another pkg version issue.

dpo:
What issue?

RenanOD:
Something with my NLPModels, julia claims there's some version requirement. I'll try reinstalling some pkgs.

dpo:
AmplNLReader should be much easier to install now.

dpo:
Will this be updated?

tmigot:
This PR has been inactive for too long. We will keep the list of problems for future implementations.

PR #50, opened 2020-12-07
use GitHub Actions

amontoison:


dpo:
The tests should exclude `template.jl`. I don't know why it wasn't a problem before.

dpo:
Great.

PR #51, opened 2021-03-19
Create CITATION.bib

tmigot:
Add bib file with the citation.

Link to Zenodo notice: https://zenodo.org/record/3672094


dpo:
Could you also add the usual notice in the README?l

dpo:
Thanks

Issue #52, opened 2021-04-12
Problems and check lists


tmigot:
I was wondering if there was a checklist of the problem sets used in `OptimizationProblems.jl`?
I noticed that most (all?) of the CUTEr/st, Schittkowski collection, and Moré-Garbow-Hillstrom have been implemented.

Having such a list would probably help to suggest new ones.

Along the same line, what would you think about a function that verifies if a problem is already on the list (maybe up to a constant scaling)? It would help to review new ones.

dpo:
Yes, a list would be great. We've been mostly working out of one of the tech reports by Luksan & Vlcek. We're not anywhere near having all CUTEst problems; there are > 1000 of them. Also, there are a bunch of problems with just one variable that I think should be removed from here.

> what would you think about a function that verifies if a problem is already on the list (maybe up to a constant scaling)? It would help to review new ones.

How would that work? Random sampling?

tmigot:
I thought we could use the `Projects` feature in Github to keep track of each list of problems (with a link and a date) and to see what is already done and in review.

Yes, mainly random sampling, something like:
- check the number of variables, number of constraints;
- randomly evaluate the objective and constraint functions;
- if we had `!=0` values in the previous step, check also if there is a constant scaling.
It is a bit naive but should be an helpful first step for reviewing even if checking only 2-3 points.

abelsiqueira:
I like the `Projects` idea. Do you permission to set it up? I personally use issues and ZenHub to keep a board, so I don't remember how `Projects` work.

About checking whether a problem is already here, one very useful thing that we couldn't make yet is something related to #3, i.e., a way to select problems.

tmigot:
> I like the `Projects` idea. Do you permission to set it up? I personally use issues and ZenHub to keep a board, so I don't remember how `Projects` work.

The potential drawback of Projects is that one needs permission to create one. Maybe `issues` are actually sufficient.

> About checking whether a problem is already here, one very useful thing that we couldn't make yet is something related to #3, i.e., a way to select problems.

That would be great indeed.

dpo:
> Yes, mainly random sampling

It's difficult because problems come in variable sizes. I think adding new problems and selecting problems conveniently should take precedence on checking whether a problem is already in there.

tmigot:
There is now a JSO-project for this. See for instance #111

Issue #19, opened 2017-06-28
Problem size uniformization


vepiteski:
When creating a problem, the first parameter should be
- always present
- the problem's dimension
Now, for instance, clplatea(1000) produces a 1 000 000 variables problem while palmer1c(1000)  crashes since no parameter is expected. This makes difficult to test batches of problems. I recommend to use the warnings as in fminsurf2 when the specified dimension is not possible and is replaced.

vepiteski:
Perhaps a verbose parameter could turn off the warnings?

dpo:
Problems that come in a fixed size should not really take a parameter. A workaround could be to declare then as `function myproblem(args...) ... end` and to ignore `args...`.

@abelsiqueira

abelsiqueira:
I like the workaround, with a recommendation that the user should check the size of the returned problem. Maybe in the README, maybe on `using OptimizationProblems`?

For a more complete solution, I think that #3 is in order. Perhaps with [NLPClass.jl](https://github.com/abelsiqueira/NLPClass.jl), but I haven't got the time to improve it.

abelsiqueira:
Proposal: Every problem must accept two positional arguments, nvar and ncon. If possible, the problem will be created with nvar variables and ncon constraints **or more** (always round up). Additional arguments should be keyword.

The default values of nvar and ncon could be the lowest possible values (nvar > 1 and ncon > 0, if present).

Related to NLSProblems.jl as well, though there we might want an additional positional argument nequ.

@Xx-fractale-xX

dpo:
In my experience, it's rare that the user can choose both nvar and ncon. Usually, there is one parameter that determines the problem "size" (nvar *and* ncon).

abelsiqueira:
So, instead, a single mandatory parameter, number of variables, and if something else can be changed it's keyword?

abelsiqueira:
A practical example:
- `arglin*` (has n and m);
- `arwhead` (has n only);
- `broydn7d` (has n and p, a parameter).

Currently:
```julia
arwhead(100) # works, n = 100
arglina(100) # works, n = 100
broydn7d(100) # works, n = 100
arwhead(100, 100) # error, only one argument is expected
arglina(100, 100) # works, n = 100, m = 100
broydn7d(100, 100) # works, n = 100, and sets the parameter, but m = 0
```

tmigot:
Can we close this issue after #90 ?

[...]

PR #56, opened 2021-04-19
Classification and meta [WIP]

tmigot:
This is the first try to make a `DataFrame` with metadata related to each problem.
Each problem would have a `Dict` containing meta information that is concatenated at pre-compilation.

Related to issues #52 #3 #19


tmigot:
How to generate the Meta:
```
using OptimizationProblems, NLPModelsJuMP

str = OptimizationProblems.generate_meta("hs116", false, false, false, false, :academic, false)
```

How to access the DataFrames with all the meta:
```
OptimizationProblems.meta
```

Issue #62, opened 2021-09-15
Add ADNLPModels test problems


tmigot:
As discussed with @dpo  and @abelsiqueira. Making submodules inside OptimizationProblems.jl for pure JuMP models, and AD problems.

- [X] Move to a folder the submodule PureJuMP
- [x] ~~Add functions to get NLPModelJuMP versions. (JuMP)~~
- [x] Make a folder for the submodule ADNLPModels
- [x] ~~Add functions for the different backends~~
- [x] Add documentation for submodules
- [x] Complete the list of AD problems
- [x] Add meta for each problem
- [x] Add a global meta
- [x] Add documentation for meta

PR #65, opened 2021-09-15
Add functions to get NLPModelJuMP versions. (JuMP submodule)

tmigot:
#62

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,10 +4,12 @@ version = "0.3.0"
 
 [deps]
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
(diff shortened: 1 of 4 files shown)


tmigot:
> Do we really need to add a dependency here? It's trivial to build the NLPModelJuMP from the pure JuMP model.

You are right, it's not a big feature. We could use `Require` to make `NLPModelsJuMP` an optional dependency or just add an example to the docs later?

dpo:
I'm not sure `Requires` would help in the case of JuMP models because it wouldn't result in any extra functionality. I suggest we simply keep exporting the pure JuMP models. It's a good idea to add an example to the docs to show to turn one into an NLPModelJuMP.

PR #66, opened 2021-09-16
Add folder and problems for ADNLPs

tmigot:
Bringing problems in ADNLPModels format #62 . These problems include a `meta` describing the problems.

Instead of using ADNLPModels directly, we could use `Requires` as the breakage tests of `ADNLPModels.jl` would be enough to see breaking releases.

145 problems from the PureJuMP submodule are missing here, and 3 problems here are not in PureJuMP.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,18 +3,18 @@ uuid = "5049e819-d29b-5fba-b941-0eee7e64c1c6"
 version = "0.3.0"
 
 [deps]
-ADNLPModels = "54578032-b7ea-4c30-94aa-7cbd1cce6c9a"
 JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 
 [compat]
-ADNLPModels = "0.3"
 JuMP = "~0.19, ~0.20, ~0.21"
 julia = "~1"
 
 [extras]
+ADNLPModels = "54578032-b7ea-4c30-94aa-7cbd1cce6c9a"
 NLPModels = "a4795742-8479-5a88-8948-cc11e1c8c1a6"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["NLPModels", "Test"]
+test = ["ADNLPModels", "NLPModels", "Test"]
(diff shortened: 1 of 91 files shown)


dpo:
> 145 problems from the PureJuMP submodule are missing here, and 3 problems here are not in PureJuMP.

Do you think you could make a list of those problems in an issue?

src/ADNLPProblems/ADNLPProblems.jl
```diff
@@ -0,0 +1,18 @@
+module ADNLPProblems
+
+using ADNLPModels, LinearAlgebra
+import ..PureJuMP
```
dpo, line 4: Why do we need PureJuMP in here?

Project.toml
```diff
@@ -3,14 +3,18 @@ uuid = "5049e819-d29b-5fba-b941-0eee7e64c1c6"
 version = "0.3.0"
 
 [deps]
+ADNLPModels = "54578032-b7ea-4c30-94aa-7cbd1cce6c9a"
```
dpo, line 6: Let's use Requires.

dpo (commented):
Thank you! How are we going to check that the implementation of those problems is correct?

src/ADNLPProblems/ADNLPProblems.jl
tmigot: Sorry for that, I removed it.

Project.toml
tmigot: > > 145 problems from the PureJuMP submodule are missing here, and 3 problems here are not in PureJuMP.
> 
> Do you think you could make a list of those problems in an issue?

No problem, I will do it as soon as this one is merged (unless you prefer to list to be completed first?)

tmigot:
> > 145 problems from the PureJuMP submodule are missing here, and 3 problems here are not in PureJuMP.
> 
> Do you think you could make a list of those problems in an issue?

No problem, I will do it as soon as this one is merged (unless you prefer to list to be completed first?)

tmigot:
> Thank you! How are we going to check that the implementation of those problems is correct?

I planned a sanity check comparing the NLPModelsJuMP version of the JuMP models, and the ADNLPModels. Similar to `test/compare_cutest.jl`. I think it can be done in a separate PR improving the unit tests.

dpo:
FreeBSD tests Julia 1.0. That should be changed to 1.3.

dpo (approved):

abelsiqueira:
Thanks

PR #69, opened 2021-09-23
Complete problem list and tests

tmigot:
- I added tests to check the implementation: NLPModelsJuMP's version vs ADNLPModels (test obj and violation of constraints) as mentioned in #66 
- completed the list of AD problems #62 
```
using OptimizationProblems
import ADNLPModels
setdiff(names(OptimizationProbnlems.PureJuMP), names(OptimizationProblems.ADNLPProblems)) # return [:PureJuMP]
```
-  A couple of problems are posing issues:
    -  :fminsrf2, :clplatea, :clplateb, :clplatec  # issue because variable is a matrix. I can make it a separate issue.
    - :hs61 #because nlpmodelsjump is not working here [see #84 in NLPModelsJuMP.jl](https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/issues/84)
- I will add the meta for the new problems in separate PR.


test/runtests.jl
```diff
@@ -1,32 +1,67 @@
...
+  @test (nlp_jump.meta.lvar == nlp_ad.meta.lvar) && (nlp_jump.meta.uvar == nlp_ad.meta.uvar)
+
+  x1 = nlp_ad.meta.x0 + rand(nlp_ad.meta.nvar)/10
+  x2 = nlp_ad.meta.x0 + rand(nlp_ad.meta.nvar)/10
+  n0 = max(abs(obj(nlp_ad, nlp_ad.meta.x0)), 1)
+  @test isapprox(obj(nlp_ad, x1), obj(nlp_jump, x1), atol=1e-14 * n0) 
+  @test isapprox(obj(nlp_ad, x2), obj(nlp_jump, x2), atol=1e-14 * n0)
+
+  for xj in [x1, x2]
+    vioad = vio(cons(nlp_ad, xj) , nlp_ad.meta.lcon, nlp_ad.meta.ucon)
```
abelsiqueira, line 38:
```suggestion
    vioad = vio(cons(nlp_ad, xj), nlp_ad.meta.lcon, nlp_ad.meta.ucon)
```

test/runtests.jl
```diff
@@ -1,32 +1,67 @@
...
   println(prob)
   prob_fn = eval(Meta.parse("OptimizationProblems.PureJuMP.$(prob)"))
-  prob_fn()
-end
+  model = prob_fn(ndef)
 
-@test OptimizationProblems.ADNLPProblems.problems == []
+  prob == :fminsrf2 && continue # issue because variable is a matrix!
+  prob == :hs61 && continue #because nlpmodelsjump is not working here https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/issues/84
+  prob in [:clplatea, :clplateb, :clplatec] && continue # issue because variable is a matrix!
```
abelsiqueira, line 22:
```suggestion
  prob == :hs61 && continue #because nlpmodelsjump is not working here https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/issues/84
  prob in [:clplatea, :clplateb, :clplatec, :fminsrf2] && continue # issue because variable is a matrix!
```

test/runtests.jl
```diff
@@ -1,32 +1,67 @@
...
 
-import ADNLPModels
+  nlp_jump = MathOptNLPModel(model)
+  nlp_ad = eval(Meta.parse("OptimizationProblems.ADNLPProblems.$(prob)()"))
 
-ndef = OptimizationProblems.ADNLPProblems.default_nvar
-for prob in OptimizationProblems.ADNLPProblems.problems
+  @test (nlp_jump.meta.nvar == nlp_ad.meta.nvar) && (nlp_jump.meta.x0 == nlp_ad.meta.x0)
+  @test nlp_jump.meta.ncon == nlp_ad.meta.ncon
+  @test (nlp_jump.meta.lvar == nlp_ad.meta.lvar) && (nlp_jump.meta.uvar == nlp_ad.meta.uvar)
```
abelsiqueira, line 29:
```suggestion
  @test nlp_jump.meta.nvar == nlp_ad.meta.nvar
  @test nlp_jump.meta.x0 == nlp_ad.meta.x0
  @test nlp_jump.meta.ncon == nlp_ad.meta.ncon
  @test nlp_jump.meta.lvar == nlp_ad.meta.lvar
  @test nlp_jump.meta.uvar == nlp_ad.meta.uvar
```
It's easier to find out where's the issue if it fails.

test/runtests.jl
```diff
@@ -1,32 +1,67 @@
...
+    vioad = vio(cons(nlp_ad, xj) , nlp_ad.meta.lcon, nlp_ad.meta.ucon)
+    vioju = vio(cons(nlp_jump, xj), nlp_jump.meta.lcon, nlp_jump.meta.ucon)
+    @test isapprox(vioad, vioju)
+  end
+end
+
+for prob in names(OptimizationProblems.ADNLPProblems)
   prob == :ADNLPProblems && continue
+  prob == :fminsrf2 && continue # issue because variable is a matrix
+  prob in [:clplatea, :clplateb, :clplatec] && continue # issue because variable is a matrix
```
abelsiqueira, line 47:
```suggestion
  prob in [:clplatea, :clplateb, :clplatec, :fminsrf2] && continue # issue because variable is a matrix
```

abelsiqueira (approved):
Thanks! I didn't review the problems, only the first files and the last. I have some small corrections, but after that, I think it's good, so I'm already marking it as approved.

test/runtests.jl
tmigot: Thanks!

test/runtests.jl
```diff
@@ -1,32 +1,67 @@
...
+  x1 = nlp_ad.meta.x0 + rand(nlp_ad.meta.nvar)/10
+  x2 = nlp_ad.meta.x0 + rand(nlp_ad.meta.nvar)/10
+  n0 = max(abs(obj(nlp_ad, nlp_ad.meta.x0)), 1)
+  @test isapprox(obj(nlp_ad, x1), obj(nlp_jump, x1), atol=1e-14 * n0) 
+  @test isapprox(obj(nlp_ad, x2), obj(nlp_jump, x2), atol=1e-14 * n0)
+
+  for xj in [x1, x2]
+    vioad = vio(cons(nlp_ad, xj), nlp_ad.meta.lcon, nlp_ad.meta.ucon)
+    vioju = vio(cons(nlp_jump, xj), nlp_jump.meta.lcon, nlp_jump.meta.ucon)
+    @test isapprox(vioad, vioju)
```
dpo, line 41: Why not just check that constraint values are the same?

tmigot: This is because the `cons` function for JuMP model is sometimes returning `c(x)-lvar` (for equality constrained problems for instance) even so it was given as a `c(x)=lvar` constraint.

dpo: But so then the `vio` test would also fail.

tmigot: If we are checking the violation of the constraint only, having `c(x) >= lvar`, or `c(x) - lvar >= 0` is the same I think so the test passes. While comparing the constraint value would give `c(x)` in one case and `c(x) - lvar` in the other case.

dpo: So for JuMP problems, c(x) = lvar becomes c(x) - lvar = 0 and lvar is set to 0 internally?

tmigot: Yes, I got problems like that.

dpo: This is another way to specify HS55:
```julia
nlp = Model()
x0 = [1, 2, 0, 0, 0, 2]
uvar = [1, Inf, Inf, 1, Inf, Inf]
@variable(nlp, 0 ≤ x[i = 1:6] ≤ uvar[i], start = x0[i])
@constraint(nlp, x[1] + 2 * x[2] + 5 * x[5] == 6)
@constraint(nlp, x[1] + x[2] + x[3] == 3)
@constraint(nlp, x[4] + x[5] + x[6] == 2)
@constraint(nlp, x[1] + x[4] == 1)
@constraint(nlp, x[2] + x[5] == 2)
@constraint(nlp, x[3] + x[6] == 2)
@NLobjective(nlp, Min, x[1] + 2 * x[2] + 4 * x[5] + exp(x[1] * x[4]))

model = MathOptNLPModel(nlp)
```
Then,
```julia
julia> model.meta.lcon
6-element Vector{Float64}:
 6.0
 3.0
 2.0
 1.0
 2.0
 2.0

julia> cons(model, zeros(model.meta.nvar))
6-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0
 0.0
 0.0

```

tmigot: Ok, thanks. I will review them to avoid the `vio` test.

tmigot: I reviewed the problems, and we now compare directly `meta.lcon`, `meta.ucon` and `cons`.

dpo: Goodness!!! That's a lot of work! Thank you!

PR #75, opened 2021-10-05
Add meta folder

tmigot:
Create the meta folder and fill-in most of the entries, using the following:
```
include("test/utils.jl")

all = setdiff(union(names(OptimizationProblems.ADNLPProblems), names(OptimizationProblems.PureJuMP)), [:PureJuMP, :ADNLPProblems])

for name in string.(all)
  io = open("src/Meta/" * name * ".jl", "w")
  write(io, generate_meta(name))
  close(io)
end
```

Next step will be to regroup all the meta in a single DataFrame.

A number of entries will have to be completed separately. Default values for undetermined entries:
- `contype`: `:general` (`:linear` if `nlp.meta.nlin == nlp.meta.ncon`, or `unconstrained`)
- `objtype`: `:other`
- `cvx_obj`: `false`
- `cvx_con`: `false` (`true` if unconstrained)
- `cq`: `UInt8(0)`
- `optimal_value`: `NaN`
- `has_multiple_solution`: `missing`
- `not_everywhere_defined`: `missing`
- `origin`: `:unknown`
- `deriv`: `UInt8(0)`

`is_infeasible` is `missing` unless the problem is unconstrained or `nlp.meta.x0` is feasible.

PR #76, opened 2021-10-29
Add scripts for meta and one example

tmigot:
@abelsiqueira The PR you suggested in #75 

I already renamed `not_everywhere_defined` as `everywhere_defined`, `has_multiple_solution` as `has_multiple_solutions`, `has_cvx_obj` as `cvx_obj` and `hax_cvx_con` as `cvx_feasible_set`.

The example was generated by
```
using ADNLPModels
using OptimizationProblems
include("test/utils.jl")
set_meta([:AMPGO02]) # requires that the folder src/Meta exists
```


tmigot:
The change in `runtests.jl` will break the coverage, as there exists an old version of the meta for some problems in the submodule ADNLPProblems. [They will have to be removed in another PR]

test/utils.jl
```diff
@@ -0,0 +1,136 @@
...
+  end
+
+  str = "$(name)_meta = Dict(
+  :nvar => $(nlp.meta.nvar),
+  :variable_size => $(variable_size),
+  :ncon => $(nlp.meta.ncon),
+  :variable_con_size => $(variable_con_size),
+  :minimize => $(nlp.meta.minimize),
+  :name => \"$(name)\",
+  :has_equalities_only => $(length(nlp.meta.jfix) == nlp.meta.ncon > 0),
```
abelsiqueira, line 62: Maybe we should have more meta information and avoid some of these `has` things. This would allow more customized filtering later.

abelsiqueira: Things like the number of equalities, inequalities, etc.

test/utils.jl
```diff
@@ -0,0 +1,136 @@
...
+  end
+end
+
+"""
+  `generate_meta(model, name)`
+  `generate_meta(name)`
+  
+  is used to generate the meta of a given NLPModel.
+
+Default values for undetermined entries:
```
abelsiqueira, line 17: We should probably add the description of these to the README (and docs?)

test/utils.jl
```diff
@@ -0,0 +1,136 @@
+using ADNLPModels, NLPModels, NLPModelsJuMP, OptimizationProblems
+
+function set_meta(all::AbstractVector{T}) where T <: Union{Symbol, String}
+  for name in string.(all)
+    io = open("src/Meta/" * name * ".jl", "w")
+    write(io, generate_meta(name))
+    close(io)
+  end
```
abelsiqueira, line 8: We should have a way of reusing the parts of the meta that have to be manually informed, though this can be done later.

test/utils.jl
```diff
@@ -0,0 +1,136 @@
...
+- `cq`: `UInt8(0)`
+- `optimal_value`: `NaN`
+- `has_multiple_solutions`: `missing`
+- `everywhere_defined`: `missing`
+- `origin`: `:unknown`
+- `deriv`: `UInt8(0)`
+`is_infeasible` is `missing` unless the problem is unconstrained or `nlp.meta.x0` is feasible.
+"""
+function generate_meta(nlp::AbstractNLPModel, name::String)
+  contype = if nlp.meta.ncon == 0 && !(length(nlp.meta.ifree) < nlp.meta.nvar)
```
abelsiqueira, line 31:
```suggestion
  contype = if nlp.meta.ncon == 0 && length(nlp.meta.ifree) >= nlp.meta.nvar
```

test/utils.jl
```diff
@@ -0,0 +1,136 @@
...
+function is_infeasible(nlp::AbstractNLPModel; x = nlp.meta.x0)
+  infeasible = missing
+  if nlp.meta.ncon == 0
+    infeasible = false
+  else
+    c = cons(nlp, x)
+    vio = max(maximum(c - nlp.meta.ucon), maximum(nlp.meta.lcon - c))
+    if vio <= 0
+      infeasible = false
+    end
```
abelsiqueira, line 106: I didn't get the idea here.

abelsiqueira (changes requested):
Hi @tmigot, thanks for the PR. I made some suggestions, let me know what you think.

test/utils.jl
tmigot: If `c - ucon <= 0` and `lcon - c < = 0` then the problem is not infeasible

test/utils.jl
tmigot: Thanks !

test/utils.jl
tmigot: Sure, I can add another function like this one that replaces the name of some entries without replacing the whole meta.

test/utils.jl
tmigot: Once the meta is added to this repo, it will be the end of a series of large changes. So, I planned to update the README and the docs at the end.

test/utils.jl
tmigot: The problem with many of the meta-information is that they aren't flexible enough for scalable problems. 
That's the idea of the `get_name_meta(::Integer)` function.

test/utils.jl
```diff
@@ -0,0 +1,136 @@
...
+
+  default_nvar = ADNLPProblems.default_nvar
+  if name in ["clplatea", "clplateb", "clplatec", "fminsrf2"] # issue because variable is a matrix
+    variable_size, variable_con_size = false, false
+    nvar_formula, ncon_formula = "$(nlp.meta.nvar)", "$(nlp.meta.ncon)" # default values, to be modified
+  else
+    variable_size, variable_con_size, nvar_formula, ncon_formula = var_size(name, default_nvar)
+  end
+
+  str = "$(name)_meta = Dict(
```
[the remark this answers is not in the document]
tmigot: Sure why not. Is there a particular reason to use `OrderedDict` here instead of `Dict`?

test/utils.jl
tmigot: `replace_entry_meta([:AMPGO02], :everywhere_defined, :defined_everywhere)`

test/utils.jl
tmigot: I already have a draft for the doc [here](https://juliasmoothoptimizers.github.io/ADNLPModelProblems.jl/dev/tutorial/)

test/utils.jl
abelsiqueira: Got it. I think it's easier to use `if all(lcon .<= c .<= ucon)`. But we should also check variable bounds before stating `infeasible = false`. Also, we can determine some infeasible problems with `any(lvar .> uvar) || any(lcon .> ucon)`.

test/utils.jl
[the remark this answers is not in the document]
abelsiqueira: Discussed.

test/utils.jl
abelsiqueira: :+1:

abelsiqueira (approved):

abelsiqueira:
Thanks! Glad to merge this.

tmigot:
#62

PR #77, opened 2021-11-01
Remove old meta

tmigot:
From a previous merge, there exists an old version of meta for some problems (86 of them).
This PR removes the old metas in view of #76 and #75 that will add them in a corrected version.


abelsiqueira:
Thanks for the PR

PR #78, opened 2021-11-10
Add JuMP 0.22 in the Project.toml

amontoison:


amontoison:
We have a small issue with `NLPModelsJuMP` and `OptimizationProblems`.
They both depend on the other one.

tmigot:
> We have a small issue with `NLPModelsJuMP` and `OptimizationProblems`. They both depend on the other one.

I don't see the dependency in `OptimizationProblems` here: https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/blob/main/Project.toml ?
I think you have to update `NLPModelsJuMP` first, make a new release and then update here, no?

amontoison:
> > We have a small issue with `NLPModelsJuMP` and `OptimizationProblems`. They both depend on the other one.
> 
> I don't see the dependency in `OptimizationProblems` here: https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/blob/main/Project.toml ? I think you have to update `NLPModelsJuMP` first, make a new release and then update here, no?

It's in a the docs/Project.toml : https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/blob/main/docs/Project.toml
Yes, I need to update `NLPModelsJuMP` first.

PR #79, opened 2021-11-10
Release 0.4.0

amontoison:


amontoison:
We need a major release because we have two collections of problems now (PureJuMP and ADNLPProblems).

tmigot:
Thanks @amontoison ! Let me just add some docs about this change, because it will be the first stable build of the docs and it is better if we have something to show there :).

PR #84, opened 2021-11-19
Add upper bound

tmigot:
@amontoison I missed that the new release failed https://github.com/JuliaRegistries/General/pull/48855
I think this PR should fix the problem.


abelsiqueira:
THanks

amontoison:
Thanks Tangi and Abel.
I will try to register the release 0.4 again.

PR #88, opened 2021-11-22
Add `default_nvar` sanity check

tmigot:


test/runtests.jl
```diff
@@ -82,3 +82,30 @@ for prob in names(ADNLPProblems)
...
+
+names_pb_vars = meta[meta.variable_nvar .== true, [:nvar, :name, :best_known_upper_bound, :best_known_lower_bound]]
+adproblems = (eval(Meta.parse("ADNLPProblems.$(pb[:name])()")) for pb in eachrow(names_pb_vars))
+for (pb, nlp) in zip(eachrow(names_pb_vars), adproblems)
+  @test pb[:nvar] == nlp.meta.nvar
+  if is_feasible(nlp) && get_minimize(nlp)
+    @test pb[:best_known_upper_bound] ≈ obj(nlp, nlp.meta.x0)
+  elseif is_feasible(nlp) && !get_minimize(nlp)
+    @test pb[:best_known_lower_bound] ≈ obj(nlp, nlp.meta.x0)
+  end
```
abelsiqueira, line 110: These will fail for manually updated problems, right?

tmigot: Ouch, thanks for catching that. That was for the local test only.

abelsiqueira:
Thanks!

PR #90, opened 2021-11-22
Uniform size `PureJuMP`

tmigot:
Use the constant `default_nvar` as a default parameter in scalable jump models 
#19


abelsiqueira:
Thank you

PR #97, opened 2021-12-02
Add contributing.md with guidelines for new problems

tmigot:
#98 can serve as an example

### DIFF
--- a/docs/src/contributing.md
+++ b/docs/src/contributing.md
@@ -1,0 +1,33 @@
+# Contributing to OptimizationProblems.jl
+
+First off, thanks for taking the time to contribute!
+
+## Bug reports and discussions
+
+If you think you found a bug, feel free to open an [issue](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/issues).
+Focused suggestions and requests can also be opened as issues. Before opening a pull request, please start an issue or a discussion.
+
+If you want to ask a question not suited for a bug report, feel free to start a discussion [here](https://github.com/JuliaSmoothOptimizers/Organization/discussions), a forum for general discussion about this repository and the [JuliaSmoothOptimizers](https://github.com/JuliaSmoothOptimizers) organization. Discussions about any of our packages are welcome.
+
+## Adding new problems
+
+We welcome pull requests proposing new problems to the problem set. As a general guideline, a pull request should concern one problem only. We recommend checking existing problems as a template for your new problems.
+
+Here is a to-do list, to help you add new problems:
+* Before implementing a new problem, make sure it does not already exist in this repository.
+* This package contains implementations using [`JuMP`](https://github.com/jump-dev/JuMP.jl) and [`ADNLPModels`](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl). A pull request should include both implementations of a new problem. Additionally, a "meta" provides general information regarding the problem. Therefore, a PR adding a new problem should contain 3 files:
+    - `src/ADNLPProblems/name_of_the_problem.jl`
+    - `src/PureJuMP/name_of_the_problem.jl`
+    - `src/Meta/name_of_the_problem.jl`
+* When submitting a problem, please pay particular attention to the documentation. We would like to gather as much information as possible on the provenance of problems, other problem sets where the problems are present, and general information on the problem. 
+The documentation should be added to the file in the `PureJuMP` folder.
+* New problems can be scalable, see [ADNLPProblems/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/arglina.jl) and [PureJuMP/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/PureJuMP/arglina.jl) for examples. In that case, the first keyword parameter should be the number of variables `n::Int` and have the default value `default_nvar` (constant predefined in the module). If your problem has restrictions on the number of variables, e.g., `n` should be odd, or `n` should have the form `4k + 3`, then, instead of throwing errors when the restrictions are not satisfied, you should instead use the number of variables to be as close to `n` as possible. For example, if you want `n` odd and `n = 100` is passed, you can internally convert to `n = 99`. If you want `n = 4k + 3`, and `n = 100` is passed, then compute `k = round(Int, (n - 3) / 4)` and update `n`.
+* A first version of the `meta` can be generated using `generate_meta`. A `String` is returned that can be copy-pasted into the `Meta` folder, and then edited.
+
+```julia
+    include("test/utils.jl")
+    #load your model in the environment
+    generate_meta(model, "name_of_the_problem")
+```
+
+* Problems modeled with `ADNLPModels` should be type-stable, i.e. they should all have keyword argument `type::Val{T} = Val(Float64)` where `T` is the type of the initial guess and the type used by the `NLPModel` API.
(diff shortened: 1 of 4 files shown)


docs/src/contributing.md
```diff
@@ -0,0 +1,33 @@
+# Contributing to OptimizationProblems.jl
+
+First off, thanks for taking the time to contribute!
+
+## Bug reports and discussions
+
+If you think you found a bug, feel free to open an [issue](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/issues).
+Focused suggestions and requests can also be opened as issues. Before opening a pull request, start an issue or a discussion on the topic, please.
```
dpo, line 8:
```suggestion
Focused suggestions and requests can also be opened as issues. Before opening a pull request, please start an issue or a discussion.
```

docs/src/contributing.md
```diff
@@ -0,0 +1,33 @@
...
+
+If you want to ask a question not suited for a bug report, feel free to start a discussion [here](https://github.com/JuliaSmoothOptimizers/Organization/discussions). This forum is for general discussion about this repository and the [JuliaSmoothOptimizers](https://github.com/JuliaSmoothOptimizers), so questions about any of our packages are welcome.
```
dpo, line 10:
```suggestion
If you want to ask a question not suited for a bug report, feel free to start a discussion [here](https://github.com/JuliaSmoothOptimizers/Organization/discussions), a forum for general discussion about this repository and the [JuliaSmoothOptimizers](https://github.com/JuliaSmoothOptimizers) organization. Discussions about any of our packages are welcome.
```

docs/src/contributing.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Adding new problems
+
+We welcome pull requests proposing new problems to the problem set. As a general guideline, a pull request should concern one problem only. We recommend checking existing problems as a template for your new problems.
+
+Here is a to-do list, to help you add new problems:
+* Before implementing a new problem, make sure it does not already exist in this repository.
+* This package contains implementations using [`JuMP`](https://github.com/jump-dev/JuMP.jl) and [`ADNLPModels`](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl). New problems should include both implementations. Additionally, a meta provides general information regarding the problem. Therefore, a PR adding a new problem should contain 3 files:
```
dpo, line 18:
```suggestion
* This package contains implementations using [`JuMP`](https://github.com/jump-dev/JuMP.jl) and [`ADNLPModels`](https://github.com/JuliaSmoothOptimizers/ADNLPModels.jl). A pull request should include both implementations of a new problem. Additionally, a "meta" provides general information regarding the problem. Therefore, a PR adding a new problem should contain 3 files:
```

docs/src/contributing.md
```diff
@@ -0,0 +1,33 @@
...
+    - `src/ADNLPProblems/name_of_the_problem.jl`
+    - `src/PureJuMP/name_of_the_problem.jl`
+    - `src/Meta/name_of_the_problem.jl`
+* When submitting a problem, please pay particular attention to the documentation. We would like to gather as much information as possible on the provenance of problems, other problem sets where the problems are present, and general information on the problem. 
+The documentation should be added to the file in the `PureJuMP` folder.
+* New problems can be scalable, see [ADNLPProblems/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/arglina.jl) and [PureJuMP/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/PureJuMP/arglina.jl) for an example. In that case, the first parameter should have the value `default_nvar` (constant predefined in the module) and be as close as possible to the number of variables.
```
dpo, line 24:
```suggestion
* New problems can be scalable, see [ADNLPProblems/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/arglina.jl) and [PureJuMP/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/PureJuMP/arglina.jl) for an example. In that case, the first parameter should have the value `default_nvar` (a constant predefined in the module) and be as close as possible to the number of variables.
```

dpo: What does "and be as close as possible to the number of variables" mean?

docs/src/contributing.md
```diff
@@ -0,0 +1,33 @@
...
+* A first version of the `meta` can be generated using `generate_meta`. A `String` is returned that can be copy-paste in the `Meta` folder, and then edited.
```
dpo, line 25:
```suggestion
* A first version of the `meta` can be generated using `generate_meta`. A `String` is returned that can be copy-pasted into the `Meta` folder, and then edited.
```

docs/src/contributing.md
```diff
@@ -0,0 +1,33 @@
...
+
+```julia
+    include("test/utils.jl")
+    #load your model in the environment
+    generate_meta(model, "name_of_the_problem")
+```
+
+* The problems modeled with `ADNLPModels` are type-stable, i.e. they all have a keyword argument `type::Val{T} = Val(Float64)` where `T` is the type of the initial guess and the type used by the `NLPModel` API.
```
dpo, line 33:
```suggestion
* Problems modeled with `ADNLPModels` should be type-stable, i.e. they should all have keyword argument `type::Val{T} = Val(Float64)` where `T` is the type of the initial guess and the type used by the `NLPModel` API.
```

dpo (changes requested):
This is great, thanks!

docs/src/contributing.md
```diff
@@ -0,0 +1,33 @@
...
+    - `src/ADNLPProblems/name_of_the_problem.jl`
+    - `src/PureJuMP/name_of_the_problem.jl`
+    - `src/Meta/name_of_the_problem.jl`
+* When submitting a problem, please pay particular attention to the documentation. We would like to gather as much information as possible on the provenance of problems, other problem sets where the problems are present, and general information on the problem. 
+The documentation should be added to the file in the `PureJuMP` folder.
+* New problems can be scalable, see [ADNLPProblems/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/arglina.jl) and [PureJuMP/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/PureJuMP/arglina.jl) for an example. In that case, the first parameter should have the value `default_nvar` (constant predefined in the module) and be as close as possible to the number of variables, i.e. the number of variables of `name_of_the_problem(n)` should be close to `n`.
```
abelsiqueira, line 24:
```suggestion
* New problems can be scalable, see [ADNLPProblems/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/arglina.jl) and [PureJuMP/arglina.jl](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/PureJuMP/arglina.jl) for a examples. In that case, the first keyword parameter should be the number of variables `n::Int` and have the default value `default_nvar` (constant predefined in the module). If your problem has restrictions on the number of variables, e.g., `n` should be odd, or `n` should have the form `4k+3`, then, instead of throwing errors when the restrictions are not satisfied, you should instead use the number of variables to be as close to `n` as possible. For example, if you want `n` odd and `n = 100` is passed, you can internally convert to `n = 99`. If you want `n=4k+3`, and `n = 100` is passed, then compute `k = round(Int, (n-3)/4)` and update `n`.
```
Also, there was a test for this that needs to be changed?

tmigot: There was no tests for that. In Issue #19 , I had a piece of code to raise warnings. Do you think we should add a warning in the unit tests in case we suspect the number of variables is "too far" from `default_nvar`?

Also, for PureJuMP model `n` is not a kwargs, but a classical args with a default value. I will probably uniformize that in a future PR.

abelsiqueira:
Thanks! Merged after meeting discussion.

Issue #111, opened 2021-12-21
Test set: Modified CUTE Problems for Sparse Unconstrained Optimization


tmigot:
Complete the list from Luksan - Matonoha - Vlček test set: http://www.cs.cas.cz/matonoha/download/V1081.pdf

Suggested in https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/issues/7

- [ ] Only problem `Problem 54. SPMSRTLS` seems to be missing.

PR #112, opened 2021-12-21
Remove constant woods and clean docs

tmigot:
A problem `woods` was already defined in `woods.jl`.

I fixed some typos related to the technical report #111

### DIFF
--- a/src/PureJuMP/chainwoo.jl
+++ b/src/PureJuMP/chainwoo.jl
@@ -19,7 +19,7 @@
 #
 # D. Orban, Montreal, 08/2015.
 
-export chainwoo, woods
+export chainwoo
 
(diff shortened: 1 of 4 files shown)


dpo:
Good catch. Do the woods problems reduce to the same actual problem?

tmigot:
Apparently, they are the same for `n < 8` according to this:
```
for n=1:100
  nlp1 = OptimizationProblems.ADNLPProblems.chainwoo(n=n)
  nlp = OptimizationProblems.ADNLPProblems.woods(n=n)
  x = rand(nlp.meta.nvar)
  if nlp1.f(x) != nlp.f(x)
    @show n, abs(nlp1.f(x) - nlp.f(x))
  end
end
```

dpo:
Looks great. Thanks!

PR #125, opened 2022-02-24
use `m` instead of `n` for ad `brownden`

tmigot:


src/PureJuMP/brownden.jl
```diff
@@ -12,23 +12,23 @@
...
-  m = max(m, 4)
+function brownden(args...; n::Int = default_nvar, kwargs...)
+  n < 4 && @warn("brownden: must have n ≥ 4")
+  n = max(n, 4)
 
   nlp = Model()
 
   x0 = [25.0, 5, -5, -1]
 
   @variable(nlp, x[j = 1:4], start = x0[j])
```
dpo, line 23: We're only setting the first 4 components of x0. What about the rest when n > 4 ?

tmigot: Oh sorry, I misunderstood the problem. After checking the paper again, the number of variables is indeed 4, but `m` control the complexity of the objective.
Actually, in the ADNLPProblems version 
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/brownden.jl
we should replace `n` by `m` to be consistent, and keep `n` for the number of variables.

tmigot:
Change `n` in `m` just to avoid any confusion with the parameter `n` that usually sets the number of variables and to be consistent with the JuMP model.

Issue #126, opened 2022-03-02
Error in AD problems


tmigot:
I did some checking on the ADNLPProblems with the following
```
using ADNLPModels, OptimizationProblems, NLPModels
ad_problems = [OptimizationProblems.ADNLPProblems.eval(Symbol(pb))() for pb in OptimizationProblems.meta[!, :name]]
for pb in ad_problems @show pb.meta.name;x = rand(pb.meta.nvar);jac(pb, x);grad(pb, x) end
```
and a number of problems returned some errors.
**Domain error**
- [ ] hs110

**Error in the jacobian computation**
```
julia> jac(nlp, x)
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{OptimizationProblems.ADNLPProblems.var"#c#929"{Float64}, Float64}, Float64, 2})
```
- [x] mgh01feas
- [x] lincon
- [x] linsv
- [x] hs10
- [x] hs11
- [x] hs14

**Error in the gradient**
```
julia> grad(nlp, x)
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{OptimizationProblems.ADNLPProblems.var"#f#627"{Float64, Vector{Float64}}, Float64}, Float64, 3})
```
- [x] hs25
- [x] AMPGO13

tmigot:
I close this as the remaining error is outside the bounds.

PR #127, opened 2022-03-02
Remove `T` to convert vector

tmigot:
#126 

```
using ADNLPModels, OptimizationProblems, NLPModels
ad_problems = [OptimizationProblems.ADNLPProblems.eval(Symbol(pb))() for pb in setdiff(OptimizationProblems.meta[!, :name], ["hs110"])]
for pb in ad_problems @show pb.meta.name;x = rand(pb.meta.nvar);jac(pb, x);grad(pb, x) end
```
now works.

### DIFF
--- a/src/ADNLPProblems/AMPGO13.jl
+++ b/src/ADNLPProblems/AMPGO13.jl
@@ -3,7 +3,7 @@ export AMPGO13
 function AMPGO13(; n::Int = default_nvar, type::Val{T} = Val(Float64), kwargs...) where {T}
   function f(x)
     n = length(x)
-    return 0 < x[1] < 1 ? -T(x[1]^(2 / 3) + (1 - x[1]^2)^(1 / 3)) : T[Inf]
+    return 0 < x[1] < 1 ? -(x[1]^T(2 / 3) + (1 - x[1]^2)^T(1 / 3)) : T[Inf]
   end
   x0 = T[1 / 1000]
   return ADNLPModels.ADNLPModel(f, x0, name = "AMPGO13"; kwargs...)
--- a/src/ADNLPProblems/AMPGO13.jl
+++ b/src/ADNLPProblems/AMPGO13.jl
@@ -3,7 +3,7 @@ export AMPGO13
 function AMPGO13(; n::Int = default_nvar, type::Val{T} = Val(Float64), kwargs...) where {T}
   function f(x)
     n = length(x)
-    return 0 < x[1] < 1 ? -(x[1]^T(2 / 3) + (1 - x[1]^2)^T(1 / 3)) : T[Inf]
+    return 0 < x[1] < 1 ? -(x[1]^(2 // 3) + (1 - x[1]^2)^(1 // 3)) : T[Inf]
   end
   x0 = T[1 / 1000]
   return ADNLPModels.ADNLPModel(f, x0, name = "AMPGO13"; kwargs...)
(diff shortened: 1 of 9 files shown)


src/ADNLPProblems/AMPGO13.jl
```diff
@@ -3,7 +3,7 @@ export AMPGO13
 function AMPGO13(; n::Int = default_nvar, type::Val{T} = Val(Float64), kwargs...) where {T}
   function f(x)
     n = length(x)
-    return 0 < x[1] < 1 ? -T(x[1]^(2 / 3) + (1 - x[1]^2)^(1 / 3)) : T[Inf]
+    return 0 < x[1] < 1 ? -(x[1]^T(2 / 3) + (1 - x[1]^2)^T(1 / 3)) : T[Inf]
```
dpo, line 6: I don't think it's necessary to enclose Ints or rationals with `T` either. Also `T[Inf]` could again cause problem. Is there some sort of `convert(eltype(x), Inf)`?

src/ADNLPProblems/hs10.jl
```diff
@@ -3,7 +3,7 @@ export hs10
 function hs10(; n::Int = default_nvar, type::Val{T} = Val(Float64), kwargs...) where {T}
   x0 = T[-10; 10]
   f(x) = x[1] - x[2]
-  c(x) = T[-3 * x[1]^2 + 2 * x[1] * x[2] - x[2]^2 + 1]
+  c(x) = [-3 * x[1]^2 + 2 * x[1] * x[2] - x[2]^2 + 1]
   lcon = T[0.0]
```
dpo, line 7: explicit Float64 here

tmigot: There is a unit test for this and the type seems correct (not restricted to Float64).

src/ADNLPProblems/AMPGO13.jl
abelsiqueira: Also `T(1 / 3)` will create Float64 1/3 and then convert to `T`, which for `BigFloat` is loss of precisions. `(T(2) / 3)` is uglier but better.

src/ADNLPProblems/hs25.jl
```diff
@@ -1,10 +1,10 @@
 export hs25
 
 function hs25(; n::Int = default_nvar, type::Val{T} = Val(Float64), kwargs...) where {T}
-  u = [25 + (-50 * log(0.01 * i))^(2 / 3) for i = 1:99]
+  u = T[25 + (-50 * log(0.01 * i))^(2 / 3) for i = 1:99]
```
abelsiqueira, line 4: Same comment as earlier

src/ADNLPProblems/AMPGO13.jl
dpo: I think we're supposed to leave it  as `2 / 3` to be generic.

abelsiqueira: What do you mean by generic?

dpo: I mean that Ints and rationals will be converted automatically to the appropriate floating-point type.

abelsiqueira: `2 / 3` is just the `Int64` division. We can use `2 // 3` instead to have Rational, but that might be slower.

dpo: No actually 2/3 is a Float64. Let's use 2//3 then.

abelsiqueira (approved):
LGTM, thanks.

PR #130, opened 2022-03-24
Fix dixmaan name

d-monnet:


src/ADNLPProblems/dixmaan_efgh.jl
```diff
@@ -21,14 +21,14 @@ function dixmaane(;
            sum(T(i / n) * T(δ) * x[i] * x[i + 2 * m] for i = 1:m)
   end
   x0 = 2 * ones(T, n)
-  return ADNLPModels.ADNLPModel(f, x0, name = "dixmaan_efgh"; kwargs...)
+  return ADNLPModels.ADNLPModel(f, x0, name = "dixmaan_e"; kwargs...)
```
dpo, line 24: Thanks. I don't think there should be an underscore though.

dpo (changes requested):

src/ADNLPProblems/dixmaan_efgh.jl
tmigot: In the `meta` https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/1ef559df2d101557f23e7eed8a2bff09c76be05a/src/Meta/dixmaane.jl#L7
there is no underscore.

d-monnet: I'll make a new pull request with no underscore

d-monnet:
New pull request with no underscore in the dixmaan problems name will be done

PR #131, opened 2022-03-24
Fix dixmaan problems name (no underscore)

d-monnet:


tmigot (approved):
I also tested locally
```
using ADNLPModels
using OptimizationProblems.ADNLPProblems
nlp = dixmaanf()
nlp.meta.name # returns "dixmaanf"
```
Thanks !!

PR #140, opened 2022-04-14
BOX3

SandraParent:
add of BOX3 problem

Issue #115, opened 2021-12-21
Moré - Garbow - Hillstrom test set


tmigot:
Moré, J. J., Garbow, B. S., & Hillstrom, K. E. (1981).
Testing Unconstrained Optimization Software. ACM Transactions on Mathematical Software, 7(1), 17–41. 
https://dl.acm.org/doi/10.1145/355934.355936

tmigot:
![chebyquad](https://user-images.githubusercontent.com/25304288/235475573-9d1bd17b-8ab4-416e-9db6-de779ce3cb84.png)

The issue is with the computation of the Chebyshev polynomials that are defined recursively. I tried things like
```
m = n = 100
function Chbi(y::T, m) where {T} # Chebyshev polynomial of the first kind
    res = zeros(T, m)
    m = Int(m)
    res[1] = 1 # T0
    if i == 1
      return res
    end
    res[2] = y
    if m == 2
      return res
    end
    for j=2:m
      res[2 + j] = 2 * y * res[j + 1] - res[j]
    end
    return res[end]
  end
  register(nlp, :Chbi, 2, Chbi, autodiff = true)
```
doesn't work because it supports only real arguments.

@amontoison any idea?

amontoison:
Sorry Tangi, I misunderstood the issue. Can you definite the function in the the function that definites the JuMP model such that you can remove the argument `m`?

tmigot:
Well, it's true we could try to simplify the `sum` in `NLobjective` but I had similar issues for other models as well and was wondering if there was a simpler solution.

dpo:
I think there’s a problem here with at least certain models of this collection. They are created as `ADNLPModel`s while they should be `ADNLSModel`s. For example, here is how `biggs6` (problem 18) is created:

https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/biggs6.jl#L15

If you try to solve this problem, you find that it’s unbounded. Of course, that doesn’t make any sense, because it should be a nonlinear least-squares objective.

Perhaps other NLS problems are affected as well.

PR #142, opened 2022-04-14
add box2 problem from cutest

tmigot:
Box2 variant of #140 from CUTEst

https://bitbucket.org/optrove/sif/src/master/BOX2.SIF


tmigot:
This is connected to https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/issues/115

Issue #161, opened 2022-04-20
COPS 3.0


tmigot:
These are large scale parametrizable problem from https://www.mcs.anl.gov/~more//cops/cops3.pdf

- [ ] Largest Small Polygon
- [ ] Distribution of Electrons on a Sphere
- [ ] Shape Optimization of a Cam
- [ ] Hanging Chain
- [ ] Isometrization of α-pinene
- [ ] Marine Population Dynamics
- [ ] Flow in a Channel
- [ ] Robot Arm
- [ ] Particle Steering
- [ ] Goddard Rocket
- [ ] Hang Glider
- [ ] Catalytic Cracking of Gas Oil
- [ ] Methanol to Hydrocarbons
- [ ] Catalyst Mixing
- [ ] Elastic-Plastic Torsion
- [ ] Journal Bearing
- [ ] Minimal Surface with Obstacle
- [ ] Triangular Mesh Smoothing
- [ ] Tetrahedral Mesh Smoothing
- [ ] Transition States for the Lane-Emden Problem
- [ ] Transition States for the Dirichlet Problem
- [ ] Transition States for the Henon Problem

nathanemac:
Hello, 
I would like to work on the following three problems: 
2- Distribution of Electrons on a Sphere
6- Marine Population Dynamics
8- Robot Arm

Thank you, 
Nathan Allaire

indyalardjane:
Hello,
I would like to work on the following three problems:
10 - Goddard Rocket
11 - Hang Glider
17 - Minimal Surface with Obstacle

Thank you, 
Indya Lardajne

MohamedLaghdafHABIBOULLAH:
Hello,
I would like to work on the following three problems:
3 - Shape Optimization of a Cam
18 - Triangular Mesh Smoothing
19 - Tetrahedral Mesh Smoothing

Best regards,
Mohamed Laghdaf HABIBOULLAH.

dpo:
Those models are available here in AMPL format: http://www.mcs.anl.gov/~more/cops/cops-3.0.zip

dpo:
Please also see https://juliasmoothoptimizers.github.io/OptimizationProblems.jl/stable/contributing/

tmigot:
I updated the list with the names in the CUTEst lib. However, I didn't find CUTEst entries for the following: bearing - dirichlet - henon - lane-hemden - tetra - torsion - triangle

[...]

PR #167, opened 2022-04-22
Add `polygon`

tmigot:
#161


tmigot:
I think we miss the initial guess for this one.

dpo:
Here's one: https://vanderbei.princeton.edu/ampl/nlmodels/polygon/polygon2.mod

tmigot:
> Here's one: https://vanderbei.princeton.edu/ampl/nlmodels/polygon/polygon2.mod

Great, thanks! I adapted it to the current model.

PR #176, opened 2022-05-13
add `hs219`

indyalardjane:


src/ADNLPProblems/hs219.jl
```diff
@@ -0,0 +1,18 @@
...
+  function c(x)
+    return [
+        x[1]^2 - x[2] - x[4]^2,
+        x[2] - x[1]^3 - x[3]^2
+    ]
+  end
+  x0 = 10 * ones(T, 4)
+  lcon = zeros(T, 2)
+  ucon = zeros(T, 2)
+  nlp = ADNLPModels.ADNLPModel(f, x0, c, lcon, ucon, name = "hs219", lin = [1]; kwargs...)
```
dpo, line 16: The first constraint isn't linear.

dpo (changes requested):
There is an error in the tests. You can see the details by clicking on the individual checks. Here it is: https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/runs/6426341685?check_suite_focus=true#step:6:325

Issue #174, opened 2022-05-12
Type unstability of obj() with IntervalBox input


d-monnet:
Issues:

1. Looks like the returned type of obj() is not the same than the second argument when this last is an IntervalBox from the IntervalArithmetics library for some problems.
2. Error indicates that ^ is not defined for Interval{32}, but it is, for :"broydn7d", "hs101", "hs102", "hs103", "hs104", "hs110", "hs25", "hs70"

I am not sure is the problem comes from JSO or IntervalArithmetics though.

Run attached code for details.
[interval_type_unstable.zip](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/files/8682788/interval_type_unstable.zip)

d-monnet:
Note: I is necessary to cast an IntervalBox into vector{Interval} to call obj(). This corresponds to the first lines of the code.

tmigot:
Thanks @d-monnet for reporting this issue.

The issue for `AMPGO14` and `schmvett` comes from our package. It misses a conversion of `pi` for both.

Then, there is an issue with the function `^` and also for `abs`. The problem seems to be that these functions are not defined for `Interval` derived type used by ForwardDiff. Maybe you can report this on `IntervalArithmetic` or `ForwardDiff` directly. For instance,
```
ForwardDiff.gradient(x -> abs(x[1]), X0)
```

tmigot:
Following #184 , it's all from IntervalArithmetics so I am closing this.

PR #184, opened 2022-05-31
Pi convertion in AMPGO14 and schmvett

d-monnet:
Convert pi into T(pi), ensures type stability of the models.

### DIFF
--- a/src/ADNLPProblems/AMPGO14.jl
+++ b/src/ADNLPProblems/AMPGO14.jl
@@ -3,7 +3,7 @@ export AMPGO14
 function AMPGO14(; n::Int = default_nvar, type::Val{T} = Val(Float64), kwargs...) where {T}
   function f(x)
     n = length(x)
-    return -exp(-x[1]) * sin(2 * (pi * x[1]))
+    return -exp(-x[1]) * sin(2 * (T(pi) * x[1]))
   end
   x0 = zeros(T, 1)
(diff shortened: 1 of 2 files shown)


src/ADNLPProblems/AMPGO14.jl
```diff
@@ -3,7 +3,7 @@ export AMPGO14
 function AMPGO14(; n::Int = default_nvar, type::Val{T} = Val(Float64), kwargs...) where {T}
   function f(x)
     n = length(x)
-    return -exp(-x[1]) * sin(2 * (pi * x[1]))
+    return -exp(-x[1]) * sin(2 * (T(pi) * x[1]))
```
dpo, line 6: Should have been π. It shouldn't be necessary to convert it. I would expect promote and conversion to work with irrationals.

d-monnet: The pull request follows up the comment  to that issue:
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/issues/174#issuecomment-1125464039

pi does not modify the type T indeed. Looks like the problem might come from JuliaInterval library.

d-monnet: The error comes from IntervalArithmetic library. Multiplying pi by an interval, whaterver its type, returns a Float64 interval.

PR #210, opened 2022-08-15
fix rosenbrock fixes #208

### LINKED ISSUE #208: Fixed `n` in `rosenbrock`
@dpo, @tmigot, I tried to call 
```julia
n=10
PureJuMP.rosenbrock(;n)
```
and I ended up with a problem with 2 variables:
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/8a219602a1357eac7ed84c632ec93204ecec8820/src/PureJuMP/genrose.jl#L77

I think `n=2` could be removed in the line
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/8a219602a1357eac7ed84c632ec93204ecec8820/src/ADNLPProblems/genrose.jl#L16

I can open a PR replacing both lines with :
```julia
 rosenbrock(args...; kwargs...) = genrose(; kwargs...) 
```
if you are ok with this change.

paraynaud:
@tmigot about ADNLPProblems.rosenbrock, fixes #208

### DIFF
--- a/src/ADNLPProblems/genrose.jl
+++ b/src/ADNLPProblems/genrose.jl
@@ -13,4 +13,4 @@ function genrose(; n::Int = default_nvar, type::Val{T} = Val(Float64), kwargs...
   return ADNLPModels.ADNLPModel(f, x0, name = "genrose"; kwargs...)
 end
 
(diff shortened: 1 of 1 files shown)


tmigot (approved):
Looks good, thanks!

Issue #213, opened 2022-09-06
StackOverFlow Error occurs when using the module with multiple processes.


MonssafToukal:
I wanted to use `OptimizationProblems.jl` to benchmark some solvers in parallel, but I get a strange error.

```
using Distributed
addprocs(2; exeflags="--project=$(@__DIR__)")

@everywhere begin
  using ADNLPModels, OptimizationProblems, OptimizationProblems.ADNLPProblems
end
```

This gives the following error: 
```
      From worker 2:    ┌ Warning: Error requiring `ADNLPModels` from `OptimizationProblems.ADNLPProblems`
      From worker 2:    │   exception =
      From worker 2:    │    LoadError: StackOverflowError:
      From worker 2:    │    Stacktrace:
      From worker 2:    │      [1] promote_typeof(::Float64, ::Float64, ::Vararg{Any}) (repeats 13902 times)
      From worker 2:    │        @ Base ./promotion.jl:330
      From worker 2:    │      [2] hvcat(::NTuple{4199, Int64}, ::Int64, ::Vararg{Number})
      From worker 2:    │        @ Base ./abstractarray.jl:2019
      From worker 2:    │      [3] top-level scope
      From worker 2:    │        @ ~/.julia/packages/OptimizationProblems/nfPUU/data/tetra_duct12.jl:1
      From worker 2:    │      [4] include(mod::Module, _path::String)
      From worker 2:    │        @ Base ./Base.jl:418
      From worker 2:    │      [5] include(x::String)
      From worker 2:    │        @ OptimizationProblems.ADNLPProblems ~/.julia/packages/OptimizationProblems/nfPUU/src/ADNLPProblems/ADNLPProblems.jl:1
      From worker 2:    │      [6] top-level scope
      From worker 2:    │        @ ~/.julia/packages/OptimizationProblems/nfPUU/src/ADNLPProblems/tetra.jl:71
      From worker 2:    │      [7] include(mod::Module, _path::String)
      From worker 2:    │        @ Base ./Base.jl:418
      From worker 2:    │      [8] include(x::String)
      From worker 2:    │        @ OptimizationProblems.ADNLPProblems ~/.julia/packages/OptimizationProblems/nfPUU/src/ADNLPProblems/ADNLPProblems.jl:1
      From worker 2:    │      [9] top-level scope
[...]

PR #226, opened 2022-11-24
Use `jld2` to store data

tmigot:
@MonssafToukal could you check if that solve https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/issues/213 ?


MonssafToukal:
This seems to have fixed it. Thanks @tmigot !

PR #249, opened 2023-04-02
Etiennne tremblay patch 1

EtiennneTremblay:
Adding the Jump Model and ADNLP Model for function 211 for MTH8408 at Polytechnique Montréal.


f_211_function_Jump_ADNLP
```diff
@@ -0,0 +1,13 @@
+# Modèle JuMP
+f_211 = Model(Ipopt.Optimizer)
```
tmigot, line 2:
```suggestion
f_211 = Model()
```

tmigot:
superseded by #346

PR #255, opened 2023-04-08
Add problems `argtrig`, `bard`, `biggs5`, `brownal` and `broyden3d`

tmigot:
#115


src/PureJuMP/bard.jl
```diff
@@ -0,0 +1,30 @@
+#   Bard problem in 3 variables.
+#   This function is a nonlinear least squares with 15 groups.
+#   Each group has a linear and a nonlinear element.
+#
+#   Source: Problem 3 in
```
tmigot, line 5:
```suggestion
#   Source: Problem 8 in
```

PR #283, opened 2023-05-22
Add catenary

MaxenceGollier:


tmigot:
Hi @MaxenceGollier ! Thank you for the pull request, before I start the review, can I ask where did you find the formulation of the problem? Is it from the original paper?

MaxenceGollier:
> Hi @MaxenceGollier ! Thank you for the pull request, before I start the review, can I ask where did you find the formulation of the problem? Is it from the original paper? 

Yes, i  originally found the problem on CUTest, i adapted it a bit as the original article was in latin ..

PR #310, opened 2023-09-23
[documentation] Update Documenter.jl

amontoison:


docs/Project.toml
```diff
@@ -10,9 +10,9 @@ Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 SolverBenchmark = "581a75fa-a23a-52d0-a590-d6201de2218a"
 
 [compat]
-ADNLPModels = "0.6"
-Documenter = "0.27"
-JSOSolvers = "0.9"
-NLPModels = "0.19"
+ADNLPModels = "0.7"
+Documenter = "1.0"
```
tmigot, line 14: Why 1.0 and not 1?

tmigot:
It looks like here too reference.md is too big. Can you at least update the packages and keep Documenter to the working version?
Thanks

docs/Project.toml
amontoison: I like to put major and minor version numbers for aesthetic reasons :)

tmigot: Ok, the advantage of putting just 1 is that we no longer have to track 1.1, 1.2, etc

amontoison: I don't think so Tangi: https://pkgdocs.julialang.org/v1/compatibility/#Caret-specifiers

tmigot: Oh, you are right, thanks for that!

PR #331, opened 2024-06-01
Add breakage tests

### LINKED ISSUE #330: Add Breakage.yml workflow with JSOSuite
Version 0.7.4 was a breaking change (changed the API) and so it broke JSOSuite.
This could have been caught if we had a breakage workflow and ran that before releasing 0.7.4

tmigot:
Close #330


abelsiqueira (approved):
Thanks! I haven't reviewed the file, I assume it was copied from someplace. But I approve of the idea.

PR #321, opened 2024-02-29
Update JuMP syntax to use the new nonlinear interface

frapac:
I had [to comment out the tests](https://github.com/frapac/OptimizationProblems.jl/blob/fp/nljump/test/runtests.jl#L47-L53) checking that the formulation in JuMP matches those in ADNLPModels.

Apparently, the problem arises when we have `ScalarQuadraticFunction` in the MOI model. Maybe related to: 
https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/pull/102

PR #334, opened 2024-06-21
Fix `hs61` test

### LINKED ISSUE #72: Fix unit tests for `hs61`
Since PR #69 , the JuMP models have an ADNLPModels translation, and we check the compatibility in the unit tests using [NLPModelsJuMP.jl](https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl).
There is an issue with `hs61` (see [runtests.jl L16](https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/c562568f35b52fd017128f8a1c5f433c01a201ef/test/runtests.jl#L16)) because MathOptNLPModel is not working in this situation [see #84 in NLPModelsJuMP.jl](https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/issues/84).

tmigot:
Close #72 
Need https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/pull/181

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -45,7 +45,6 @@ include("test_utils.jl")
   end
 
   @testset "Test problems compatibility for $prob" begin
-    prob == :hs61 && continue #because nlpmodelsjump is not working here https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/issues/84
     prob_fn = eval(Meta.parse("PureJuMP.$(prob)"))
     model = prob_fn(n = ndef)
     nlp_jump = MathOptNLPModel(model)


amontoison:
I already updated `hs61` a few weeks ago.
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/commit/f01b2918506e520caa0e4f54095da37fa6d8c4b8

amontoison:
We need to do that only if we want to switch to the new API. It's related to the PR #321.

tmigot:
Yes and no, using @constraint for quadratic constraint is possibly a better modeling, the fact that use @NLconstraint was a hack. I don't think this is particularly connected to the new API.
Thanks to the great work in NLPModelsJuMP we will be able to update quadratic constraints, that was the point of this. Also this PR was a draft, there was no need to merge it.

amontoison:
@NLconstraint was mandatory for quadratic constraints before MOI, it was not a hack.
JuMP just added support for it later and us now.

The new API is now doing the same thing but for all nonlinear constraints /objective.

I merged the PR becasue we can test `hs61` in all cases now, even with the old API.

PR #339, opened 2024-09-01
Update JuMP syntax to use the new nonlinear interface (new)

tmigot:
Supersed #321 

Since, we test that problems are the "same" than JuMP, it imposed some reorganization on the ADNLPModels side.

- [x] hs87 is blocking https://github.com/JuliaSmoothOptimizers/NLPModelsJuMP.jl/issues/195
- [x] we need to find the match in britgas (or a work around)


test/runtests.jl
```diff
@@ -45,6 +44,8 @@ include("test_utils.jl")
   end
 
   @testset "Test problems compatibility for $prob" begin
+    pb == "britgas" && continue # the same up to some permutation of the constraints
+    pb == "hs87" && continue # if/else issue
```
tmigot, line 48:
```suggestion
```

Issue #356, opened 2024-12-18
Update names of `errinros_mod`, `chnrosnb_mod` and `fletcbv3_mod`?


danphenderson:
The current names of these ADNLPModels are `errinros`, `chnrosnb` and `fletcbv3`; respectively.

In my use case, I am instantiating a set of ADNLPModels from a list of their names; but I run into issues with these models as their names do not correspond to the name of the file in the source code.

A fix would be to update these model's names to `errinros_mod`, `chnrosnb_mod` and `fletcbv3_mod`.

danphenderson:
Also, `indef_mod`.

tmigot:
Thanks @danphenderson for the issue. You are right, bwe should modify the problems' names.

For errinros_mod:
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/f563bb655758d816eecff69eeddba7a7c866e267/src/ADNLPProblems/errinros_mod.jl#L21
and also the NLS variant:
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/f563bb655758d816eecff69eeddba7a7c866e267/src/ADNLPProblems/errinros_mod.jl#L40

This could also be tested, probably here:
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/f563bb655758d816eecff69eeddba7a7c866e267/test/runtests.jl#L29
`@test nlp_ad.meta.name == String(pb)`

strangely enough the name was correct in the dataframe describing the problems
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/f563bb655758d816eecff69eeddba7a7c866e267/src/Meta/errinros_mod.jl#L7

@danphenderson Would you be interested in opening a PR to fix this?

danphenderson:
@tmigot, thanks for the additional instructions - PR submitted.

PR #357, opened 2024-12-20
fix: discrepancy in problem names.

danphenderson:
Reference #356 

Q: How should I handle the NLS name declared here:
https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/59bf416346611e4713d1c49b9c748ad474b9fa91/src/ADNLPProblems/errinros_mod.jl#L40
should it be `errinros_mod_nls` or `errinros-mod-nls` or `errinros_mod-nls`?


tmigot:
It should be -nls

Thanks!

tmigot:
So, the tests do not pass because of rosenbrock that we define via genrose here https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src%2FADNLPProblems%2Fgenrose.jl

tmigot (approved):
LGTM, thanks!

PR #360, opened 2024-12-26
Fix small dimension specification issue.

### LINKED ISSUE #355: Fix: `clplatea` small dimension issue
Currently, ` prob = ADNLPProblems.clplatea(n=N)` fails for all $N < 9$ as a result of this sum in it's definition:

```julia
           sum(
             sum(
               1 // 2 * (x[i + (j - 1) * p] - x[i - 1 + (j - 1) * p])^2 +
               hp2 * (x[i + (j - 1) * p] - x[i - 1 + (j - 1) * p])^4 for j = 2:p
             ) for i = 3:p
           )
```

Where $p$ is defined to be $p = floor(Int, sqrt(n))$. So, $N$ must be greater than  $8$ for `sum( .... for i=3:p)` to be over a non-empty collection.

Suggested fixes:
- Update the warning statement and properly adjust the number of variables to be 9 or greater.
- Or, depending on the province and if it makes sense, we may modify the above sum as `sum((... for i = 3:p); init=0.0)`; but then an issue would arise for $N < 4$ if similar action isn't taken to modify the two other sums in `clplatea`'s definition.

danphenderson:
Resolves #355; tests are passing locally.

### DIFF
--- a/src/ADNLPProblems/clplatea.jl
+++ b/src/ADNLPProblems/clplatea.jl
@@ -6,8 +6,8 @@ function clplatea(;
   wght = -0.1,
   kwargs...,
 ) where {T}
-  p = floor(Int, sqrt(n))
-  p * p != n && @warn("clplatea: number of variables adjusted from $n down to $(p*p)")
+  p = floor(Int, sqrt(n)) < 3 ? 3 : floor(Int, sqrt(n))
+  p * p != n && @warn("clplatea: number of variables adjusted from $n to $(p*p)")
   n = p * p
   hp2 = (1 // 2) * p^2
   function f(x; p = p, hp2 = hp2, wght = wght)
--- a/src/ADNLPProblems/clplatea.jl
+++ b/src/ADNLPProblems/clplatea.jl
@@ -6,7 +6,7 @@ function clplatea(;
   wght = -0.1,
   kwargs...,
 ) where {T}
-  p = floor(Int, sqrt(n)) < 3 ? 3 : floor(Int, sqrt(n))
+  p = max(floor(Int, sqrt(n)), 3)
   p * p != n && @warn("clplatea: number of variables adjusted from $n to $(p*p)")
   n = p * p
   hp2 = (1 // 2) * p^2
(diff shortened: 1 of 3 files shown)


tmigot (approved):
Thanks @danphenderson !

src/ADNLPProblems/clplatea.jl
```diff
@@ -6,8 +6,8 @@ function clplatea(;
   wght = -0.1,
   kwargs...,
 ) where {T}
-  p = floor(Int, sqrt(n))
-  p * p != n && @warn("clplatea: number of variables adjusted from $n down to $(p*p)")
+  p = floor(Int, sqrt(n)) < 3 ? 3 : floor(Int, sqrt(n))
```
tmigot, line 9:
```suggestion
  p = max(floor(Int, sqrt(n)), 3)
```

src/ADNLPProblems/clplateb.jl
```diff
@@ -6,8 +6,8 @@ function clplateb(;
   wght = -0.1,
   kwargs...,
 ) where {T}
-  p = floor(Int, sqrt(n))
-  p * p != n && @warn("clplateb: number of variables adjusted from $n down to $(p*p)")
+  p = floor(Int, sqrt(n)) < 3 ? 3 : floor(Int, sqrt(n))
```
tmigot, line 9:
```suggestion
  p = max(floor(Int, sqrt(n)), 3)
```

src/ADNLPProblems/clplatec.jl
```diff
@@ -8,8 +8,8 @@ function clplatec(;
   l = 0.01,
   kwargs...,
 ) where {T}
-  p = floor(Int, sqrt(n))
-  p * p != n && @warn("clplatec: number of variables adjusted from $n down to $(p*p)")
+  p = floor(Int, sqrt(n)) < 3 ? 3 : floor(Int, sqrt(n))
```
tmigot, line 11:
```suggestion
  p = max(floor(Int, sqrt(n)), 3)
```

tmigot (changes requested):

PR #406, opened 2026-02-27
Add warnings when dimension is modified

### LINKED ISSUE #354: Fix: Warn when Dimension is Modified
Warn on invalid dimension for `NZF1`: ref #123 
- When a user specifies `n < 26` they should receive a warning that the dimension is being adjusted. (As done with the dixmaan* problems)

Figure out what is going on with `spmsrtls`:  It appears to only like dimension's greater than 100.. but the specified dimension is being modified slightly; users should be warned.

arnavk23:
Fixes #354

Applied the macro to all problems with automatic dimension adjustments:

| Problem | Constraint | Variants |
|---------|-----------|----------|
| **NZF1** | Multiple of 13 | `:nlp`, `:nls` |
| **spmsrtls** | Custom formula `m*3-2` | `:nlp`, `:nls` |
| **chainwoo** | Multiple of 4 | `:nlp`, `:nls` |
| **woods** | Multiple of 4 | Single |
| **srosenbr** | Multiple of 2 (even) | Single |
| **catenary** | Multiple of 3, minimum 6 | Single |
| **clplatea** | Perfect square | Single |
| **clplateb** | Perfect square | Single |
| **clplatec** | Perfect square | Single |
| **fminsrf2** | Minimum 4, then perfect square | Single |
| **powellsg** | Multiple of 4 | `:nlp`, `:nls` |
| **watson** | Clamped [2, 31] | `:nlp`, `:nls` |
| **bearing** | Grid dimensions validation | Single |


tmigot (changes requested):

tmigot:
@arnavk23 can you have a look at copilot's comments ? I think most of them are justified;

arnavk23:
@tmigot I agree. Added them all.

src/PureJuMP/PureJuMP.jl
```diff
@@ -20,6 +20,22 @@ end
 
 using JuMP, LinearAlgebra, SpecialFunctions
 
+_adjust_nvar_warn_message(problem_name, n_orig, n) =
+  string(problem_name, ": number of variables adjusted from ", n_orig, " to ", n)
```
tmigot: Is there a point having a separate function for this?

src/PureJuMP/PureJuMP.jl
```diff
@@ -20,6 +20,22 @@ end
...
+
+"""
+    @adjust_nvar_warn(problem_name, n_orig, n)
+
+Issue a warning if the number of variables was adjusted, showing both original and adjusted values.
+"""
+macro adjust_nvar_warn(problem_name, n_orig, n)
```
tmigot: Why do we have the macro in PureJuMP and in ADNLPProblems ? Can't we unify ?

tmigot (changes requested):

src/PureJuMP/broydn7d.jl
```diff
@@ -46,9 +46,10 @@ export broydn7d
 
 "Broyden 7-diagonal model in size `n`"
 function broydn7d(args...; n::Int = default_nvar, p::Float64 = 7 / 3, kwargs...)
-  mod(n, 2) > 0 && @warn("broydn7d: number of variables adjusted to be even")
+  n_orig = n
   n2 = max(1, div(n, 2))
   n = 2 * n2
+  @adjust_nvar_warn("broydn7d", n_orig, n)
```
tmigot: broydn7d is not modified in ADNLPProblems ?

src/PureJuMP/dixmaan_efgh.jl
```diff
@@ -33,9 +33,10 @@ function dixmaane(
   δ::Float64 = 0.125,
   kwargs...,
 )
-  (n % 3 == 0) || @warn("dixmaan: number of variables adjusted to be a multiple of 3")
+  n_orig = n
   m = max(1, div(n, 3))
   n = 3 * m
+  @adjust_nvar_warn("dixmaan", n_orig, n)
```
tmigot: None of the dixman is updated in ADNLPProblems ?

src/ADNLPProblems/spmsrtls.jl
```diff
@@ -59,8 +61,10 @@ function spmsrtls(::Val{:nlp}; n::Int = default_nvar, type::Type{T} = Float64, k
 end
 
 function spmsrtls(::Val{:nls}; n::Int = default_nvar, type::Type{T} = Float64, kwargs...) where {T}
+  n_orig = n
   m = max(Int(round((n + 2) / 3)), 34)
   n = m * 3 - 2
+  @adjust_nvar_warn("spmsrtls", n_orig, n)
```
tmigot: Not updated in PureJuMP?

src/ADNLPProblems/watson.jl
```diff
@@ -31,7 +33,9 @@ function watson(::Val{:nlp}; n::Int = default_nvar, type::Type{T} = Float64, kwa
 end
 
 function watson(::Val{:nls}; n::Int = default_nvar, type::Type{T} = Float64, kwargs...) where {T}
+  n_orig = n
   n = min(max(n, 2), 31)
+  @adjust_nvar_warn("watson", n_orig, n)
```
tmigot: Not updated in PureJuMP?

test/test-defined-problems.jl
```diff
@@ -22,6 +22,130 @@ probes = @sync begin
...
+function _check_adjusted_warning(ctor::Function, expected_msg::AbstractString, expected_nvar::Integer)
+  nlp = @test_logs (:warn, expected_msg) ctor()
+  @test nlp.meta.nvar == expected_nvar
+end
+
+function _check_adjusted_warning(expected_msg::AbstractString, expected_nvar::Integer, ctor::Function)
+  _check_adjusted_warning(ctor, expected_msg, expected_nvar)
+end
+
+@testset "Adjusted dimension warnings" begin
```
tmigot, line 25: Minimally it should be a loop here instead of listing problems

tmigot (changes requested):

test/test-defined-problems.jl
```diff
@@ -22,6 +22,63 @@ probes = @sync begin
...
 @info "PureJuMP missing per worker" probes
 
+function _check_adjusted_warning(ctor::Function, expected_msg::AbstractString, expected_nvar::Integer)
+  nlp = @test_logs (:warn, expected_msg) ctor()
+  @test nlp.meta.nvar == expected_nvar
+end
+
+function _check_adjusted_warning(expected_msg::AbstractString, expected_nvar::Integer, ctor::Function)
+  _check_adjusted_warning(ctor, expected_msg, expected_nvar)
+end
```
tmigot: I don't see the interest in defining these two intermediary functions?

src/OptimizationProblems.jl
```diff
@@ -2,6 +2,23 @@ module OptimizationProblems
 
 using DataFrames
 
+_adjust_nvar_warn_message(problem_name, n_orig, n) =
+  string(problem_name, ": number of variables adjusted from ", n_orig, " to ", n)
```
tmigot: Is there really a point in having this external to the macro code?

test/test-defined-problems.jl
```diff
@@ -22,6 +22,63 @@ probes = @sync begin
...
+
+@testset "Adjusted dimension warnings" begin
```
tmigot: To be honest, I don't like this test. I think it's overly complicated to maintain compare to what it test. Moreover, it doesn't even test if some problems do not show the message while they should or the inverse.
Couldn't we think of a better test?

tmigot (changes requested):

test/test-defined-problems.jl
arnavk23: I am keeping this open but I have changed the file. Can you check it out whenever you are free.

test/test-defined-problems.jl
```diff
@@ -22,6 +22,68 @@ probes = @sync begin
 end
 @info "PureJuMP missing per worker" probes
 
+function _warning_problems()
```
tmigot: Thanks for updating this. I still have some reservations about this test.

My main concern is that it discovers problems by scanning the source for `@adjust_nvar_warn`, so it mostly tests whether problems already using the macro emit the expected warning. It would not catch the opposite failure mode: a scalable problem adjusting `n` silently because the macro/warning was forgotten.

I also find the loop over hard-coded values of `n` a bit fragile. Some problems may only adjust dimensions for specific divisibility constraints, and returning after the first adjusted case means we only test one incidental example per problem.

I think a more robust approach would be to drive the test from the problem registry / scalable problem metadata instead of the implementation detail. Ideally, for each scalable problem and a few representative requested dimensions, we would instantiate the model, compare requested `n` with `meta.nvar`, and assert:
- if `meta.nvar != n`, the standardized warning is emitted;
- if `meta.nvar == n`, no warning is emitted.

That would test the actual public behavior we care about and would also catch missing warnings.

tmigot: Something like this:
```
@testset "Adjusted dimension warnings" begin
  n = 100
  var_problems = OptimizationProblems.meta[
    OptimizationProblems.meta.variable_nvar,
    :name,
  ]

  for prob in var_problems
    get_nvar = getfield(OptimizationProblems, Symbol("get_", prob, "_nvar"))
    n_adjusted = get_nvar(n)

    for mod in (ADNLPProblems, PureJuMP)
      isdefined(mod, prob) || continue

      constructor = getfield(mod, prob)

      if n_adjusted != n
        msg_re = Regex("number of variables adjusted from $(n) to $(n_adjusted)")
        @test_logs (:warn, msg_re) constructor(; n = n)
      else
        @test_nowarn constructor(; n = n)
      end
    end
  end
end
```
maybe add the content in a function and test this in another loop on the problems.

tmigot (changes requested):

test/test-defined-problems.jl
```diff
@@ -22,6 +22,65 @@ probes = @sync begin
...
+  for subdir in ("ADNLPProblems", "PureJuMP")
+    for file in readdir(joinpath(src_root, subdir))
+      endswith(file, ".jl") || continue
+      source = read(joinpath(src_root, subdir, file), String)
+      if occursin("@adjust_nvar_warn", source)
+        stem = first(splitext(file))
+        push!(probs_with_macro, stem)
+      end
+    end
+  end
```
tmigot: You don't want to do that, because you also want to catch problems that should use the macro but doesn't.

test/test-defined-problems.jl
```diff
@@ -22,6 +22,65 @@ probes = @sync begin
...
+
+  # Test each problem that uses the macro
```
tmigot: None of the comments line are useful here

test/test-defined-problems.jl
```diff
@@ -22,6 +22,65 @@ probes = @sync begin
...
+        isdefined(mod, prob_sym) || continue
+
+        constructor = getfield(mod, prob_sym)
+        
+        try
+          # Try to verify the model can be constructed with adjusted size
+          _ = constructor(; n = n_adjusted)
+        catch
+          continue  # Skip if construction fails
+        end
```
tmigot: No try/catch here, here are fully controlling the environment and it risks hidding a bug.

tmigot (changes requested):
Thanks @arnavk23 for the changes. I think we are getting there :).

arnavk23:
@tmigot Tangi other than the codecov, everything passes. Also I was able to pick up on more that needed the macro using the test file, so added them. I think this is fully done now. Thank you for all the reviews.

src/ADNLPProblems/elec.jl
```diff
@@ -1,36 +1,39 @@
...
+  n = 3 * m
+  @adjust_nvar_warn("elec", n_orig, n)
   # Define the objective function to minimize
   function f(x; n = n)
     return sum(
       sum(
-        1 / sqrt((x[j] - x[i])^2 + (x[n + j] - x[n + i])^2 + (x[2n + j] - x[2n + i])^2) for
-        j = (i + 1):n
-      ) for i = 1:(n - 1)
+        1 / sqrt((x[j] - x[i])^2 + (x[m + j] - x[m + i])^2 + (x[2m + j] - x[2m + i])^2) for
```
tmigot: @arnavk23 This PR is just supposed to add one-line with the new macro and should absolutely not modify the code of the problems.

tmigot (changes requested):

src/ADNLPProblems/elec.jl
arnavk23: Fixed everything. There shouldn't be any problem now. Wishing for the best.

src/PureJuMP/torsion.jl
```diff
@@ -7,9 +7,12 @@ export torsion
 
 function torsion(args...; n = default_nvar, kwargs...)
   # number of variables is (nx + 1) x (ny + 1)
+  n_orig = n
   if !((:nx in keys(kwargs)) & (:ny in keys(kwargs)))
-    nx, ny = Int(round(sqrt(max(1, n - 2)))), Int(round(sqrt(max(1, n - 2))))
+    nx, ny = Int(round(sqrt(max(1, n_orig - 2)))), Int(round(sqrt(max(1, n_orig - 2))))
```
tmigot:
```suggestion
    nx, ny = Int(round(sqrt(max(1, n - 2)))), Int(round(sqrt(max(1, n - 2))))
```

src/PureJuMP/torsion.jl
```diff
@@ -7,9 +7,12 @@ export torsion
...
   end
+  nvars = (nx + 2) * (ny + 2)
+  @adjust_nvar_warn("torsion", n_orig, nvars)
```
tmigot:
```suggestion
  @adjust_nvar_warn("torsion", n_orig, (nx + 2) * (ny + 2))
```

src/PureJuMP/steering.jl
```diff
@@ -7,7 +7,10 @@
 export steering
 
 function steering(; n::Int = default_nvar, kwargs...)
+  n_orig = n
```
tmigot:
```suggestion
```

src/PureJuMP/steering.jl
```diff
@@ -7,7 +7,10 @@
...
   a = 100.0  # Magnitude of force.
+  nvars = 5 * n_orig + 6
+  @adjust_nvar_warn("steering", n_orig, nvars)
```
tmigot:
```suggestion
  @adjust_nvar_warn("steering", n, 5 * n_orig + 6)
```

src/PureJuMP/rocket.jl
```diff
@@ -7,7 +7,10 @@
 export rocket
 
 function rocket(; n::Int = default_nvar, kwargs...)
+  n_orig = n
```
tmigot:
```suggestion
```

src/PureJuMP/rocket.jl
```diff
@@ -7,7 +7,10 @@
...
   h_0 = 1.0
+  nvars = 4 * n_orig + 5
+  @adjust_nvar_warn("rocket", n_orig, nvars)
```
tmigot:
```suggestion
  @adjust_nvar_warn("rocket", n, 4 * n + 5)
```

src/PureJuMP/pinene.jl
```diff
@@ -25,6 +26,8 @@ function pinene(; n::Int = default_nvar, kwargs...)
   # times at which observations made
   tau = [1230.0, 3060.0, 4920.0, 7800.0, 10680.0, 15030.0, 22620.0, 36420.0]
   tf = tau[nm]                       # ODEs defined in [0,tf]
+  n = n_orig
+  @adjust_nvar_warn("pinene", n_orig, 50 * n + 5)
```
tmigot: `n_orig = n` and then `n = n_orig` this is unnecessary

src/PureJuMP/minsurf.jl
```diff
@@ -12,10 +12,13 @@
...
 
 function minsurf(args...; n = default_nvar, kwargs...)
+  n_orig = n
   # number of variables is (nx + 2) x (ny + 2)
   if !((:nx in keys(kwargs)) & (:ny in keys(kwargs)))
-    nx, ny = Int(round(sqrt(max(1, n - 2)))), Int(round(sqrt(max(1, n - 2))))
+    nx, ny = Int(round(sqrt(max(1, n_orig - 2)))), Int(round(sqrt(max(1, n_orig - 2))))
   end
+  n = (nx + 2) * (ny + 2)
+  @adjust_nvar_warn("minsurf", n_orig, n)
```
tmigot: n_orig is useless here

src/PureJuMP/methanol.jl
```diff
@@ -35,6 +36,8 @@ function methanol(args...; n::Int = default_nvar, kwargs...)
     1.122,
   ]
   tf = tau[nm]                   # ODEs defined in [0,tf]
+  n = n_orig
```
tmigot: again n_orig = n and then n = n_orig

src/PureJuMP/elec.jl
```diff
@@ -11,7 +11,10 @@
 export elec
 
 function elec(args...; n::Int = default_nvar, kwargs...)
-  n = max(2, div(n, 3))
+  n_orig = n
+  m = max(2, div(n_orig, 3))
```
tmigot: why adding a new variable m?

test/test-defined-problems.jl
```diff
@@ -22,6 +22,34 @@ probes = @sync begin
 end
 @info "PureJuMP missing per worker" probes
 
+@testset "Adjusted dimension warnings" begin
+  var_probs = OptimizationProblems.meta[OptimizationProblems.meta.variable_nvar, :name]
+  @test !isempty(var_probs)
```
tmigot:
```suggestion
```

test/test-defined-problems.jl
```diff
@@ -22,6 +22,34 @@ probes = @sync begin
...
+      n_adjusted == n && continue  # Skip if no adjustment for this n
+
+      msg_re = Regex("number of variables adjusted from $(n) to $(n_adjusted)")
+
+      for mod in (ADNLPProblems, PureJuMP)
+        isdefined(mod, prob_sym) || continue
+
+        constructor = getfield(mod, prob_sym)
+
+            @test_logs (:warn, msg_re) constructor(; n = n)
```
tmigot:
```suggestion
        isdefined(mod, prob_sym) || continue
        constructor = getfield(mod, prob_sym)
        @test_logs (:warn, msg_re) constructor(; n = n)
```

test/test-defined-problems.jl
```diff
@@ -22,6 +22,34 @@ probes = @sync begin
...
+@testset "Adjusted dimension warnings" begin
+  var_probs = OptimizationProblems.meta[OptimizationProblems.meta.variable_nvar, :name]
+  @test !isempty(var_probs)
+
+  for prob_name in var_probs
+    prob_sym = Symbol(prob_name)
+    
+    get_nvar_func = getfield(OptimizationProblems, Symbol("get_", prob_name, "_nvar"))
+
+    for n in (50, 100, 150)
```
tmigot:
```suggestion
    for n in (50, 100)
```

tmigot (changes requested):
Hi @arnavk23 ! I checked most of the problems. That looks overall good, but several times you add new temporary variables that are useless. Please let's try to be as concise as possible.
My list of comments is not exhaustive. Thanks !!

src/ADNLPProblems/broydn7d.jl
```diff
@@ -1,9 +1,9 @@
 export broydn7d
 
 function broydn7d(; n::Int = default_nvar, type::Type{T} = Float64, kwargs...) where {T}
-  mod(n, 2) > 0 && @warn("broydn7d: number of variables adjusted to be even")
   n2 = max(1, div(n, 2))
   n = 2 * n2
+  @adjust_nvar_warn("broydn7d", n, n)
```
tmigot: ??

tmigot (commented):
What is the point of this test?

src/ADNLPProblems/broydn7d.jl
arnavk23: Fixed this. Incorrectly removed n_org here. Thanks for pointing it out.

src/ADNLPProblems/fminsrf2.jl
```diff
@@ -1,12 +1,12 @@
...
 
 function fminsrf2(; n::Int = default_nvar, type::Type{T} = Float64, kwargs...) where {T}
-  n < 4 && @warn("fminsrf2: number of variables must be ≥ 4")
   n = max(4, n)
+  n_org = n
 
   p = floor(Int, sqrt(n))
-  p * p != n && @warn("fminsrf2: number of variables adjusted from $n down to $(p*p)")
   n = p * p
+  @adjust_nvar_warn("fminsrf2", n_org, n)
```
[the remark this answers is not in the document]
tmigot: @arnavk23

src/PureJuMP/fminsrf2.jl
```diff
@@ -21,12 +21,12 @@
...
 
 function fminsrf2(args...; n::Int = default_nvar, kwargs...)
-  n < 4 && @warn("fminsrf2: number of variables must be ≥ 4")
   n = max(4, n)
+  n_org = n
 
   p = floor(Int, sqrt(n))
-  p * p != n && @warn("fminsrf2: number of variables adjusted from $n down to $(p*p)")
   n = p * p
+  @adjust_nvar_warn("fminsrf2", n_org, n)
```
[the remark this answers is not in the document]
tmigot: @arnavk23

tmigot (approved):
Thanks ! Copilot raised an issue that already existed so this PR just highlight it and I will open an issue for that.

PR #422, opened 2026-06-04
Removing public `m` from PureJuMP wrappers

### LINKED ISSUE #421: Implementation difference arglina, arglinb, arglinc
In the JuMP implementation there is a parameter `m` that doesn't exist in the ADNLPProblem implementation. those should be unified.

arnavk23:
@tmigot Fixes #421


tmigot (approved):
Thanks!

PR #426, opened 2026-06-14
Replace `Integer` with `Int` in Meta files

### LINKED ISSUE #420: Unify problem's size type
For scalable problems, we can pass as argument the desirable size.
Right now, there is an inconsistency in the typing:
- `Int` https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/ADNLPProblems/arglina.jl
- `Integer` https://github.com/JuliaSmoothOptimizers/OptimizationProblems.jl/blob/main/src/Meta/arglina.jl

And all connected ressources should be updated.

arnavk23:
Fixes #420
@tmigot done using
```
dir = "src/Meta"
for f in readdir(dir, join=true)
    if endswith(f, ".jl")
        s = read(f, String)
        s2 = replace(s, "n::Integer" => "n::Int")
        if s != s2
            println("Updating $f")
            write(f, s2)
        end
    end
end
```


tmigot (approved):
LGTM, thanks!

